PR #28, opened 2017-05-10
Why doesn't this work? Attempt multiple build targets

Entroper:


Entroper:
Ported to .NET Core 2.0, this is no longer necessary.

Issue #31, opened 2017-06-08
Investigate Early Ordeals


Entroper:
Seed 610646C1_qfLfAQKCoo, Zelly reports Early Ordeals didn't work.

Entroper:
Yeahhhhhhhhhhhhhhh that's broken.

PR #32, opened 2017-06-10
[#31] Fix early ordeals flag not being used.

bedmonds:
The early-access to the Castle of Ordeals was, incorrectly, a part of the Shuffle Ordeals flag. As a result, the Early Ordeals convenience was useless, and one could always access the dungeon early if its teleporters were being shuffled.

### DIFF
--- a/FF1Lib/FF1Rom.cs
+++ b/FF1Lib/FF1Rom.cs
@@ -81,6 +81,10 @@ namespace FF1Lib
 				ShuffleOrdeals(rng);
 			}
 
+      if (flags.EarlyOrdeals) {
+        EnableEarlyOrdeals();
+      }
+
 			if (flags.EarlyRod)
 			{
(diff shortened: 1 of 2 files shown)


Entroper:
OMG, the spacing in Maps.cs is redonkulous.  Not your fault, it was already like that.  I'll add an .editorconfig in a future commit.

This looks great BTW.  I had thought it would require another `ReadMaps/WriteMaps` call, but this just modifies the tileset, stored separately.

FF1Lib/FF1Rom.cs
```diff
@@ -81,6 +81,10 @@ public void Randomize(Blob seed, Flags flags)
 				ShuffleOrdeals(rng);
 			}
 
+      if (flags.EarlyOrdeals) {
```
Entroper: REEEEEEEEEEEEE

PR #43, opened 2017-11-09
Npc item shuffle

dhmoeller:
Integration with online/console app and flags is still needed, but this is mostly ready for testing.


dhmoeller:
This needed some cleaning up so I will resubmit a PR after making some changes.

PR #48, opened 2017-12-04
Gain multiple levels at once added to speed hacks.

BrianCumminger:
This patch allows multiple levels to be gained at the end of combat.  Due to the size of the new code required, part of the stat increase message display routine had to be overwritten and therefore this is patch also skips stat messages other than max HP.  Commented ASM source for the patch is as follows:
```
JSR $9F57 ;undraw all battle boxes
PHA ;maybe not needed
LDA $80 ;check if pointer to exp was changed
AND #$7
CMP #$7
BEQ LevelLoop
LDA $80 ;reset exp pointer
AND #$F0
ADC #$7
STA $80
LevelLoop:
PLA 
JSR $9C01 ;get exp to lvlup
JMP $9BE8
```

### DIFF
--- a/FF1Lib/Hacks.cs
+++ b/FF1Lib/Hacks.cs
@@ -86,6 +86,9 @@ namespace FF1Lib
 			Data[0x33CCD] = 0x04; // Explosion effect count (small enemies), default 8
 			Data[0x33DAA] = 0x04; // Explosion effect count (mixed enemies), default 15
 
+			//Gain multiple levels at once.  Also supresses stat increase messages as a side effect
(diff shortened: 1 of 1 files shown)


FF1Lib/Hacks.cs
```diff
@@ -86,6 +86,9 @@ public void EnableSpeedHacks()
 			Data[0x33CCD] = 0x04; // Explosion effect count (small enemies), default 8
 			Data[0x33DAA] = 0x04; // Explosion effect count (mixed enemies), default 15
 
+			//Game multiple levels at once.  Also supresses stat increase messages as a side effect
```
latzio: Gain*

BrianCumminger:
Thanks, yeah I was working on this for a while and submitted it right before falling asleep.  I probably should have known better :)

PR #51, opened 2017-12-09
Credits!

latzio:
Named a few people. Text subject to change obviously. Assert if we
go over 0x100 bytes because we'll smash into the scaled prices.

### DIFF
--- a/FF1Lib/Hacks.cs
+++ b/FF1Lib/Hacks.cs
@@ -170,7 +170,7 @@ namespace FF1Lib
 			// leading spaces are used to increment the PPU ptr precisely to save ROM space.
 			List<string[]> texts = new List <string[]>();
 			texts.Add(new string[] { "", "",
-				                     " Final Fantasy  ",
+				                     " Final  Fantasy ",
 				                     "", "",
 				                     "   Randomizer   ",
 			});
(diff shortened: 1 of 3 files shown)


FF1Lib/Hacks.cs
```diff
@@ -170,7 +170,7 @@ public void RollCredits()
 			// leading spaces are used to increment the PPU ptr precisely to save ROM space.
 			List<string[]> texts = new List <string[]>();
 			texts.Add(new string[] { "", "",
-				                     " Final Fantasy  ",
+				                     " Final  Fantasy ",
```
latzio: The extra space in between "centers" the entire string better and looks a lot better in game IMHO.

FF1Lib/Hacks.cs
```diff
@@ -185,23 +185,19 @@ public void RollCredits()
...
 									 "fcoughlin, Disch",
 									 "Paulygon, anomie",
 									 "Derangedsquirrel",
-									 "AstralEsper,    ",
+									 "AstralEsper, and",
 									 "",
 									 " The Entire FFR ",
 									 "    Community   ",
 			});
-			texts.Add(new string[] { "", "",
```
latzio: Yeah FUCK you

PR #53, opened 2017-12-17
Fix TextToBytes.

latzio:
It didn't keep a space for the null terminator, and it didn't give
you the choice of if you wanted it, so you never really knew if you
were going to get 0, 1, or many trailing nulls in your Blob.


Entroper:
We fixed this in a slightly different way.

PR #59, opened 2017-12-27
Flags Refactor and Bug Fixes

latzio:
Fixed a bug where TeamSteak flag and PartyLeaderShuffle were getting
mixed up, but also overhauled flags so they are not limited to 32 bits
anymore.

This is a little heavier than just a bugfix but it also handles growing the bitset beyond 32 bits


latzio:
Latest version highlights:
- Fixed Early Orders label click on website
- Version bump to 1.6.1 due to flags visual changes
- Reorder Team STEAK pulldown to after Party Leader checkbox. All checkboxes must be before all non-checkboxes due to continuing javascript limitations and convenience.
- Kept our zany base64/DTE/file path hybrid format and in fact added a "= to ." conversion so we can keep the base64 = signs to be more parser friendly.
 - Tested with native app and website, pasting flags into website and verifying with hex editor that rom outputs and filenames are identical.

FF1RandomizerOnline/wwwroot/js/site.js
```diff
@@ -146,56 +151,65 @@ function forcedPartyMembersCallback() {
...
-	flagsString += base64Chars[charBits];
-	charBits = (checkboxBits & 0x00000F00) >>>  6 | (checkboxBits & 0x00C00000) >>> 22;
-	flagsString += base64Chars[charBits];
-	charBits = (checkboxBits & 0x003F0000) >>> 16;
-	flagsString += base64Chars[charBits];
-	charBits = (checkboxBits & 0xFC000000) >>> 26;
-	flagsString += base64Chars[charBits];
-	charBits = (checkboxBits & 0x03000000) >>> 20;
-	flagsString += base64Chars[charBits];
+	var flagsString = btoa(String.fromCharCode.apply(null, flagBytes));
```
latzio: All the work of this patch is to leverage btoa here and atob later.
The javascript string replaces use regexes so they behave like replaceAll

PR #67, opened 2018-01-03
Add an import function, supporting crim_bot generated seed/flag strings.

bedmonds:
Because I'm salty at running the wrong seed.


Entroper:
I merged this, but quickly realized it was against master instead of dev.  The merge has been undone, please reopen against dev.

PR #69, opened 2018-01-03
Add import function, supporting native's copy/paste syntax to WWW.

bedmonds:
This one is not only against dev, it also updates the sliders. \o/

### DIFF
--- a/FF1RandomizerOnline/wwwroot/js/site.js
+++ b/FF1RandomizerOnline/wwwroot/js/site.js
@@ -31,7 +31,7 @@ function importSeedFlags(e) {
 	document.getElementById("Flags").value = flags;
 	document.getElementById("Seed").value = seed;
 
-	OnFlagsChanged();
+	onFlagsChanged();
 }
 
 function newSeed() {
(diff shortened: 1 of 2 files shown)


FF1RandomizerOnline/wwwroot/js/site.js
```diff
@@ -117,6 +121,17 @@ function setCallbacks() {
 	};
 }
 
+function OnFlagsChanged() {
```
Entroper: Make this camelCase and we're good to go.

bedmonds: We should all agree that kebab-case in a sea of parentheses is the superior choice and be done with it. :D

FF1RandomizerOnline/wwwroot/js/site.js
```diff
@@ -19,6 +19,19 @@ function validateFlags() {
...
+	var str = prompt("Paste in a seed and flags string as given to you by our lord and master, crim_bot. (SEED_FLAGS)");
+	var seed;
+	var flags;
+
+	[seed, flags] = str.split("_", 2);
+
+	document.getElementById("Flags").value = flags;
+	document.getElementById("Seed").value = seed;
+
+	OnFlagsChanged();
```
Entroper: D'oh.

bedmonds: Nerf 6:30 AM late-night coding.

PR #89, opened 2018-01-29
Further enhancements to the ModernBattlefield.

latzio:


latzio:
Still playtesting. I just need to make sure the Y changes to space the characters don't bork when they're dead or turned to stone. Everything else is pretty low risk.

PR #105, opened 2018-03-04
Fixes Enemy AoE Spells to Affect the Caster

Steven-Biro:
Overwrites the code that checks and skips casting if the target is the caster


FF1Lib/Hacks.cs
```diff
@@ -15,72 +15,72 @@ public partial class FF1Rom : NesRom
 		public const int PartyShuffleSize = 3;
```
Steven-Biro: Accidentally added this file and it isn't needed, unsure how to remove it, the updates are all line endings

Steven-Biro:
This patch overwrites the following:

    LDA $6BCE                   ; BUGGED - see if this enemy is the attacker, and skip it if it is.
    CMP btl_attacker            ;  This doesn't make any sense, as there's no reason why "target all" spells
    BEQ @Next                   ;  should skip the caster.  It doesn't behave this way for players.

FF1Lib/BugFixes.cs
```diff
@@ -73,5 +73,11 @@ public void FixEnemyStatusAttackBug()
 		{
 			Put(0x32812, Blob.FromHex("DF")); // This is the craziest freaking patch ever, man.
 		}
+
+		public void FixEnmeyAOESpells()
```
latzio: Gotta typo here in Enemy

FF1Lib/FF1Rom.cs
```diff
@@ -108,6 +108,11 @@ public void Randomize(Blob seed, Flags flags)
 				FixSpellBugs();
 			}
 
+			if (flags.EnemySpellsTargetingAllies)
+			{
+				FixEnmeyAOESpells();
```
latzio: Typo here too naturally.

FF1Lib/FF1Rom.cs
```diff
@@ -86,17 +86,17 @@ public void Randomize(Blob seed, Flags flags)
 			var rng = new MT19337(BitConverter.ToUInt32(seed, 0));
 
 			UpgradeToMMC3();
-			EasterEggs();
```
latzio: There are a lot of unnecessary whitespace changes in this file... here and several other places.

FF1Lib/Flags.cs
```diff
@@ -1,8 +1,8 @@
-﻿using System;
+﻿using System;
```
latzio: Something is really up with all the whitespace changes your editor is making.

Steven-Biro:
Fixed the problems  @latzio  pointed out

Steven-Biro:
Solves issue #105

Steven-Biro:
Included in #120

Issue #100, opened 2018-02-16
Fix enemy resistances in battle for magic


Entroper:
Enemies don't keep track of their resistances in battle.  This means XFER doesn't work on enemies, nor does AFIR work when the WzVAMP casts it.

http://www.romhacking.net/forum/index.php?topic=24130.0

PR #115, opened 2018-03-08
Puts enemy elemental resistances in ram and allows spells to affect it

Steven-Biro:
This fixes spells like XFER which affect enemy elemental resistances since they are now stored in ram, this along with my enemy AoE spells PR will allow AFIR to work on the caster.

this solves #100


Steven-Biro:
included in #120

PR #119, opened 2018-03-16
Use MD5 so the flag hash is idempotent.

latzio:
Added a dictionary in FF1Text to map the hex string to the equipment symbols.


latzio:
This is antiquated - we can now just print the whole original string.

PR #138, opened 2018-04-11
Synced windows app with new flags, fixed EncodeFlagsText bitfield issues.

roydmerkel:
Hey all,
   So I went ahead and fixed the compilation issues with the windows app (some of the texts had changed names, but were never updated.)
  While fixing this issue, I went ahead and updated the flags in the windows app to match those in the online app.
  And, while I did that, I also found that some flags weren't reencoded correctly for the call to Randomize, these are also fixed for bit field flags (I didn't delve into issues with fun% flags, I suspect I just need to add bit field data on those flags.)
  All three of these issues have been fixed.
Thanks again, Roy.


roydmerkel:
Correct, I cast it to int because some of the values were double and were barfing. I suspect this is because some of the values were double because you guys used doubles for the EXP and GOLD multipliers.

Let me noodle the second one to see if there's a consistent way to do it. We might need to do a Convert.Int32(Round(...)), or something.

Let me try to fix the fun % flags while I'm at it.

roydmerkel:
Hey, @dhmoeller , I went ahead and resubmitted with a new change, that fixes the fun % serialization/deserialization. I also traded the EXP Bonus flag to a double, leaving the ForcedPartyMembers and Music flags as integers still.

I decided to get arround this by removing the multipler flag on those two attributes and not having the code do the multiplication in that case. This is because I thought it might be confusing to cast the Music enum and ForcedPartyMembers integers from doubles back into integers in code (they have to be integers for the assembly hacks involving them, as best as I can tell -- let me know if I read the code wrong there.)

FF1Lib/Flags.cs
```diff
@@ -194,14 +195,19 @@ public class Flags : IIncentiveFlags, IMapEditFlags
...
 		public double ExpMultiplier { get; set; }
 		[FlagString(Character = 20, Multiplier = 10)]
-		public int ExpBonus { get; set; }
-		[FlagString(Character = 21, Multiplier = 1)]
+		public double ExpBonus { get; set; }
+		[FlagString(Character = 21)]
 		public int ForcedPartyMembers { get; set; }
-		
+
+		[FlagString(Character = FUN_PERCENT, FlagBit = 1)]
```
latzio: We don't want the fun% stuff in the flags string because they don't affect the game and thus different people can set them differently and still have the same experience.

roydmerkel:
Ok, removed unwated fun % saving. Fun % is now handled by storing the Flags struct directly in the MainWindow class, as a private variable, and changing it directly, instead of deserializing at the end, as was done before.

roydmerkel:
@dhmoeller , I think I found a way to turn everything from int flags into doubles for consistency. It looks like the only remaining int flag, ForcedPartyMembers, was calculated, as an int, in only one place. This meant I could do a Convert.Int32, at that spot, without issue.

The other int flag was a fun flag, and as per, @latzio , we won't be serializing it, or de-serializing it into the FlagsTextBox.Text property.

PR #186, opened 2018-07-15
Display critical hit count on critical hits

splitpunched:
No longer calls the critical hit message and instead uses the unformatted combat box buffer to create [x] crit(s)! in the combat box.


FF1Lib/FF1Rom.cs
```diff
@@ -527,7 +527,7 @@ private void ExtraTrackingAndInitCode()
 
 			// Critical hit display
 			PutInBank(0x0C, 0xA94B, Blob.FromHex("206AC2EAEA"));
```
latzio: You need to call the new entry too here 206BC2

PR #187, opened 2018-07-17
Non-incentivized shop fix.

latzio:
The shop item was being removed in the wrong spot, and potentially
twice, but not when it came from the treasurePool. That's actually
when the failure occured because there would be too many items and
not enough locations.


Entroper:
If this affects 2.3.1, go ahead and open it against `master` and I'll do a 2.3.2 tomorrow.

PR #195, opened 2018-08-19
Caravan item price based on location

splitpunched:
Removes the price reduction for the CRYSTAL and replaces it with scaling the price of the incentive item, based on wherever the item is. It defaults to Coneria being 1/8th of normal price, and Pravoka being 1/2, but this is changable.

I haven't implemented it as a flag since the Crystal being 1/8th price wasn't implemented as a flag, and I don't see "making a seed take forever because you have to grind Ogres or some shit" as an enjoyable option.

### DIFF
--- a/FF1Lib/Scale.cs
+++ b/FF1Lib/Scale.cs
@@ -49,13 +49,11 @@ namespace FF1Lib
 				prices[i] = (ushort) (flags.WrapPriceOverflow ? ((newPrice - 1) % 0xFFFF) + 1 : Min(newPrice, 0xFFFF));
             }
 			var questItemPrice = prices[(int)Item.Bottle];
-
-			// do this before the item location check fucks everything to shit
+			// If we don't do this before checking for the item shop location factor, Ribbons and Shirts will end up being really cheap
+			// This realistically doesn't matter without Shop Wares shuffle on because nobody wants to sell Ribbons/Shirts, but if it is...
 			prices[(int)Item.WhiteShirt] = (ushort)(questItemPrice / 2);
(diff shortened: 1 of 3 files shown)


FF1Lib/Scale.cs
```diff
@@ -48,17 +48,25 @@ public void ScalePrices(IScaleFlags flags, Blob[] text, MT19337 rng)
...
 			prices[(int)Item.BlackShirt] = (ushort)(questItemPrice / 2);
 			prices[(int)Item.Ribbon] = questItemPrice;
-            // Crystal can block Ship in early game where 50000 G would be too expensive
-            prices[(int)Item.Crystal] = (ushort)(prices[(int)Item.Crystal] / 8);
-
+			var itemShopFactor = new Dictionary<MapLocation, int>() {
+				{ MapLocation.Coneria, 8 },
+				{ MapLocation.Pravoka, 2 }
+			};
+			var divisor = new int();
```
latzio: I still don't think you need this line - can you do out int divisor in the following TryGetValue?

FF1Lib/Scale.cs
```diff
@@ -48,17 +48,25 @@ public void ScalePrices(IScaleFlags flags, Blob[] text, MT19337 rng)
...
+			if (itemShopFactor.TryGetValue(shopItemLocation.MapLocation, out divisor))
+			{
+				questItemPrice = (ushort)(prices[(int)Item.Bottle] / divisor);
+			}
+			for (var i = 0; i < (int)Item.Tent; i++)
+            {
```
latzio: What's up with the bracket indentation here?

PR #197, opened 2018-08-21
Exit and Warp boss tamed

Steven-Biro:
Press B now works on these spells to go back.

Probable areas for bugs to occur if they exist, when naming characters, when using warp, and when using exit.


Steven-Biro:
cherry-picked

PR #215, opened 2018-09-30
change DoPartyGen_OnCharacter, PtyGen_DrawOneText, update comments

drcatdoctor:
example info for septimus, probably not staying as-is

### DIFF
--- a/FF1Lib/asm/1E.asm
+++ b/FF1Lib/asm/1E.asm
@@ -1,37 +1,50 @@
-﻿#BANK_THIS = $1E
+﻿.include "variables.inc"
+.include "Constants.inc"
+.include "extern.inc"
+
+.export NewGamePartyGeneration
+
+.segment "ENTIRE_BANK"
+
+BANK_THIS = $1E
+
+.list on
+.listbytes unlimited
 
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 ;;
-;;  NewGamePartyGeneration  [$8000 :: 0x78010]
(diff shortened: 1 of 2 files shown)


FF1Lib/asm/1E.asm
```diff
@@ -164,66 +185,61 @@ DoPartyGen_OnCharacter:
...
+     @retry:
+        LDA ptygen_class,X         ; A = class id.  0=FI, 1=TH, BB, RM, WM, BM, FF=None
+        CLC
+        ADC #1
+        CMP #6
+        BNE :+
+          LDA #$FF
+      : STA ptygen_class, X
+        TAY
+        LDA lut_ClassMask,Y
```
Steven-Biro: Y can be 0-5 inclusive, and FF, if Y is FF, it does not load the correct value, for example from the fceux debugger
`1E:80FF:B9 15 81  LDA $8115,Y @ $8214 = #$83`
this should be loading address $8114

drcatdoctor: I did test this, but maybe I changed it during testing and checked in the wrong stuff. It was many weeks ago.

So what is supposed to be going on is that Y==FF results in LDA lut_ClassMask-1, but obviously that isn't happening. This only works if lut_ClassMask is on the zero page, and then you have to use op B5 instead of op B9.

Anyway, this is fixable with adding INY between TAY and LDA, and then moving all the lut items up one. Less cool though.

Steven-Biro:
Contained in commit "Dr. Cat's PartyGen changes"

PR #229, opened 2018-11-17
Add enum with magic names and use them for ItemMagic

nic0lette:
This adds an enum that contains all magic in the game, and then
uses that list in ItemMagic for excluding out of battle magic.

I'm planning to add a flag to refine (at least for me) which magics are used for ItemMagic, but this felt like a good first step that works regardless of the next set of changes.

If you'd rather have a PR with this + that, let me know and I'll just amend this one in the next day or so.

### DIFF
--- a/FF1Lib/ItemMagic.cs
+++ b/FF1Lib/ItemMagic.cs
@@ -1,8 +1,6 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.Linq;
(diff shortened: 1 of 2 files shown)


nic0lette:
Upon second thought, I don't think this is the right way to do this, given other stuff that's on the horizon, so I'm going to close this PR :)

PR #230, opened 2018-11-17
Minor updates to ItemMagic code.

nic0lette:
Mostly this just removes hardcoding in magic which are out-of-comat
only (such as WARP, LIFE, SOFT, etc...) and filters the list based
on the magic's properties.

Also a small bit of cleanup to have one method to produce a list of
magic spells (in Magic.cs) that's used by both it and ItemMagic.cs.


latzio:
Put in alpha instead.

PR #231, opened 2018-11-19
change Item Magic targeting to be always on

Steven-Biro:


FF1Lib/ItemMagic.cs
```diff
@@ -25,7 +25,7 @@ public partial class FF1Rom : NesRom
 
 		public void ShuffleItemMagic(MT19337 rng)
 		{
-			CastableItemTargeting(); // make items able to target a single enemy or party member
+			//CastableItemTargeting(); // make items able to target a single enemy or party member
```
latzio: Can you just delete this line entirely and the whitespace below it? I like the change because it streamlines things anyway and having big hunks of code conditionally enabled is asking for maintenance hell anyway.

PR #233, opened 2018-11-24
Add flag to allow houses to fill HP.

nic0lette:
Changes houses to restore the party's HP to full rather than just
120 HP/house. There's a minor change to the UX because of it,
but probably fine.

Specifically, rather than showing the screen with a small party
window with HP across, it simply says, "HP and magic recovered."


nic0lette:
I reverted the flag renumbering so this should have less of an effect overall. :)

PR #237, opened 2018-12-13
Add an option to improve trapped chests

nic0lette:
This adds an option to shuffle 'notable treasure' into a subset of
trapped treasure chests. The list of both the items and locations is a
best first attempt, but should be refined with more
experience/playtesting.

### DIFF
--- a/FF1Lib/ItemPlacement.cs
+++ b/FF1Lib/ItemPlacement.cs
@@ -207,19 +207,18 @@ namespace FF1Lib
 
 				// We sort the treasure pool based on value (sort of) and pull out the highest ranked ones to put
 				// in the trap chests we picked out.
-				var notableTreasurePool = treasurePool.FindAll(item => notableTreasure.Contains(item));
-				notableTreasurePool.Shuffle(rng);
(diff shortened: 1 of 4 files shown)


FF1Lib/ItemPlacement.cs
```diff
@@ -175,6 +193,47 @@ public List<IRewardSource> PlaceSaneItems(MT19337 rng)
...
+					.Concat(ItemLists.LegendaryWeaponTier)
+					.Concat(ItemLists.LegendaryArmorTier)
+					.Concat(ItemLists.RareWeaponTier)
+					.Concat(ItemLists.RareArmorTier);
+				// Convert the list to a HashSet since we'll be doing lookups in it.
+				var notableTreasure = new HashSet<Item>(notableTreasureList);
+
+				// We sort the treasure pool based on value (sort of) and pull out the highest ranked ones to put
+				// in the trap chests we picked out.
+				var notableTreasurePool = treasurePool.FindAll(item => notableTreasure.Contains(item));
```
latzio: You might want .Where instead of .FindAll but I'm honestly not sure what the difference is. My hunch would be to use .Where and ToList() at the end, and then use spliceRandom() later instead of First() and Remove() (since it doesn't both sorta)

PR #244, opened 2019-01-06
Make normal NPCs (including bats) disappear after talking

dhmoeller:
This simply makes the Talk_Norm NPC routine redirect to the routine for the disappearing guy in Castle of Ordeals. No flag added yet so it's just always on.

A more complicated approach would be to change the jump table for only the bats to redirect to the disappearing routine, but I figure for any NPC that says exactly one thing it is safe to make them vanish.

### DIFF
--- a/FF1Lib/FF1Rom.cs
+++ b/FF1Lib/FF1Rom.cs
@@ -99,6 +99,8 @@ namespace FF1Lib
 			var maps = ReadMaps();
 			var shopItemLocation = ItemLocations.CaravanItemShop1;
 
+			EnableNPCSwatter();
+
 			/*
 			flags.FreeAirship = true;
 			flags.ExperimentalFloorGeneration = true;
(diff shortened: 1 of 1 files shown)


nic0lette:
I wonder how difficult it would be to, rather than changing the code at "Talk_norm", to add a new small routine that calls Talk_Norm followed by HideThisMapObject, and then updating lut_MapObjTalkJumpTbl so only bats use it?

If other people think that hiding any NPC is fine (or better), then that's totally cool, but I guess I may feel a bit bad chatting with Arylon and having her vanish. =(

dhmoeller:
Essentially that is exactly what Talk_CoOGuy routine already does. It wouldn’t be difficult at all to simply update the jump table on a case by case basis to point bats and some other NPCs to that existing routine instead of Talk_Norm - this way was just very quick and easy to get it into alpha. 

With a bit more time, I’d like to refactor all the NPC routines, mapobjdata, and the jump table so that we can simply update the 4 data bytes in tmp 0-3 to control the NPC behavior.

latzio:
This is in, as experimental. It doesnt quite work since you can swat the LUTE plate...

PR #247, opened 2019-01-09
Add JavaScript code for WarMECH Unleashed.

nic0lette:
A single line change to support including "Unleashed" as an option for
presets.


nic0lette:
This was merged with the other PR. :)

PR #254, opened 2019-02-07
Automatically pad out seeds that are <8 digits.

nic0lette:


FF1Blazorizer/Pages/Randomize.cshtml
```diff
@@ -474,13 +474,21 @@
...
+		}
+
 		_seed = seed;
 		try
 		{
 			Blob.FromHex(_seed);
-
-			_seedInputClass = _seed.Length == 8 ? "" : "is-error";
-
+			if (seed.Length < 8)
```
latzio: I think strings have a PadLeft method you can use to just pad with whatever to a given size. Look elsewhere because I think I do it someplace else in the file already.

nic0lette: Done.

PR #257, opened 2019-02-12
What if EVERYTHING was unrunnable?

jomazeika:


FF1Blazorizer/Pages/Randomize.cshtml
```diff
@@ -77,6 +77,7 @@
 						<CheckBox Indent IsEnabled="@Flags.EnemyScripts" Id="allowUnsafePiratesCheckBox" bind-Value="@Flags.AllowUnsafePirates">Allow Unsafe Pirates</CheckBox>
 						<CheckBox Id="enemySkillsSpellsCheckBox" bind-Value="@Flags.EnemySkillsSpells">Enemy Skills/Spells</CheckBox>
 						<CheckBox Id="enemyStatusAttacksCheckBox" bind-Value="@Flags.EnemyStatusAttacks">Enemy Status Attacks</CheckBox>
+						<CheckBox Id="everythingUnrunnableCheckBox" bind-Value="@Flags.EverythingUnrunnable">All Encounters are Unrunnable</CheckBox>
```
latzio: Should this be an Indented subfield under unrunnable shuffle that's only enabled if Unrunnable shuffle is checked off?

FF1Lib/FF1Rom.cs
```diff
@@ -290,6 +290,8 @@ public void Randomize(Blob seed, Flags flags, Preferences preferences)
 				ShuffleUnrunnable(rng);
 			}
 
+			CompletelyUnrunnable();
```
latzio: You'd better check your flag here chief - and Ideally I'd actually put this inside the unrunnable shuffle block to be double obvious. 

You have to check Unrunnable Shuffle and THEN check off Completely unrunnable....

But maybe this should even be scalable? So you can have half the number of unrunnables... or double.... or 0% or 100%.... just dreaming this is a nice start.

PR #269, opened 2019-04-30
Overworld / Inn Save changes and Formation Randomizer

one1fineday:
Here are the changes to disable overworld saving with tents/cabins/houses, and to disable saving at inns.  Neither option changes the text displayed on screen.

I also have uploaded the crude version of my formation randomizer from December, but it needs a lot more work to be something worth playing.  As far as I know, the only feature that conflicts with formation randomizer is Zozo Melmond (need to define a zone for the town, and tell the standard Zozo Melmond feature to not mess with what Enemizer does there).


latzio:
I merged it into alpha instead of dev, closing.

PR #283, opened 2019-06-14
Free Ship & Separation of Chaos Rush and Treasure Hunt

stellamazeika:
Two small changes that have been tested, well-received (and would be really nice for the Onrac Summer League)


Steven-Biro:
included in the beta update

PR #286, opened 2019-07-05
Fix disabled tri-state event handling.

latzio:


FF1Blazorizer/Shared/TriStateCheckBox.cshtml
```diff
@@ -13,70 +13,76 @@
...
+	private string ValueClass => checked0 ? "tri-yes" : (checked1 ? "tri-no" : "tri-maybe");
 
-[Parameter]
-private Action<bool?> ValueChanged { get; set; }
-private bool checked0 => Value ?? false;
-private bool checked1 => !Value ?? false;
-private bool checked2 => Value == null;
+	[Parameter]
+	private Action<bool?> ValueChanged { get; set; }
+	private bool checked0 => !Disabled && (Value ?? false);
```
Steven-Biro: If someone imports flags with a disabled tri-state flags on, will it show up correctly? I dont think it should with this change

Steven-Biro: I can test it out tomorrow

latzio: That's exactly how it should work right? If shops is off, random is definitely off, right?

Steven-Biro: so long as our logic in FF1Rom.cs always matches up with the website's UI. One case where it doesnt is with Floor Shuffle.

This isnt changing the vale of Random Weapons and Armor, its only not showing you that it is set to random

Steven-Biro: If we could sync up that logic everywhere (Floor shuffle is the only one I know about), then this would be alright I think

latzio: Right we sort of did that before with greying out - that's what I need. A case filter on the image to just grey it out while preserving what it's supposed to look like.

Steven-Biro: css filter?
 filter: grayscale(100%);

latzio:
Updated, should be good now.

PR #287, opened 2019-07-15
Nitz

Entroper:
Steven seems to think you're hot on my trail.  :P


latzio:
Could you break this PR up into a few more patches? It's a little big as is to review.

Entroper:
Man, the effort to funny ratio was already kinda suffering here.  :)

PR #288, opened 2019-07-18
Enemizer update

one1fineday:
This is the enemizer update that finally, well, enemizes.  It adds the option to create totally new enemies.
I also included a fix for the Zozo Melmond issue that occurs in the current alpha, and enabling any part of Enemizer will disable shuffling formations, shuffling trap tiles, and fun% enemy names / palette swap.  (Three of these are effectively moot with Enemizer on, and the other is only marginally different and would require me to write branching paths in ShuffleTrapTiles for basically everything which I really don't want to do.)

I've completely re-did the way formations are generated, which should produce far fewer (if any) crashes, since I programmed it in a way to be a lot more tolerant and to always have some way to generate a formation (most crashes were caused by a list having nothing to select from with PickRandom or Shuffle, which is something I've consciously avoided in the current version).  Additionally, there should be no possibility whatsoever of drawing an inappropriate monster (boss or fiend) for any random formation.

The only issues I have found so far is that sometimes the formation generator creates a duplicate formation (this is something I can resolve but I'd have to go back and check every formation for duplicates and store information in memory).  There is also a potential issue with two distinct enemies of the same image sharing a palette, which I have encountered once.  I don't know the source of the issue yet but I can figure it out if I see it again and inspect the ROM.  None of these issues cause a fatal crash or anything other than mild annoyance.

### DIFF
--- a/FF1Blazorizer/Pages/Randomize.cshtml
+++ b/FF1Blazorizer/Pages/Randomize.cshtml
@@ -100,8 +100,8 @@
 							<TriStateCheckBox UpdateToolTip="@UpdateToolTipID" Indent IsEnabled="@Flags.EnemyFormationsUnrunnable" Id="everythingUnrunnableCheckBox" bind-Value="@Flags.EverythingUnrunnable">All Encounters are Unrunnable</TriStateCheckBox>
 							<TriStateCheckBox UpdateToolTip="@UpdateToolTipID" Id="unrunnablesStrikeFirstAndSurpriseCheckBox" bind-Value="@Flags.UnrunnablesStrikeFirstAndSurprise">Unrunnable First Strike &amp; Surprise</TriStateCheckBox>
 							<TriStateCheckBox UpdateToolTip="@UpdateToolTipID" Id="enemyFormationsSurpriseCheckBox" bind-Value="@Flags.EnemyFormationsSurprise">Enemy Surprise Bonus</TriStateCheckBox>
-							<EnumDropDown UpdateToolTip="@UpdateToolTipID" TItem="FormationShuffleMode" bind-Value="@Flags.FormationShuffleMode">Enemy Formations:</EnumDropDown>
-							<TriStateCheckBox UpdateToolTip="@UpdateToolTipID" Id="enemyTrapTilesCheckBox" bind-Value="@Flags.EnemyTrapTiles">Enemy Forced Encounter Tiles</TriStateCheckBox>
+							<EnumDropDown UpdateToolTip="@UpdateToolTipID" TItem="FormationShuffleMode" IsEnabled="@(!Flags.RandomizeFormationEnemizer)" bind-Value="@Flags.FormationShuffleMode">Enemy Formations:</EnumDropDown>
+							<TriStateCheckBox UpdateToolTip="@UpdateToolTipID" Id="enemyTrapTilesCheckBox" IsEnabled="@(!Flags.RandomizeFormationEnemizer)" bind-Value="@Flags.EnemyTrapTiles">Enemy Forced Encounter Tiles</TriStateCheckBox>
 							<TriStateCheckBox UpdateToolTip="@UpdateToolTipID" Indent Id="randomTrapFormationsCheckBox" IsEnabled="@Flags.EnemyTrapTiles" bind-Value="@Flags.RandomTrapFormations">Use Random Formations</TriStateCheckBox>
 							<div class="checkbox-cell"></div>
 							<h4>WarMECH</h4>
@@ -463,8 +463,8 @@
 							<div class="checkbox-cell"></div>
 							<CheckBox UpdateToolTip="@UpdateToolTipID" Id="DisableTentSaving" bind-Value="@Flags.DisableTentSaving">Disable Tent/Cabin/House Saving on Overworld</CheckBox>
 							<CheckBox UpdateToolTip="@UpdateToolTipID" Id="DisableInnSaving" bind-Value="@Flags.DisableInnSaving">Disable Inn Saving</CheckBox>
-							<CheckBox UpdateToolTip="@UpdateToolTipID" Id="RandomizeEnemizer" bind-Value="@Flags.RandomizeEnemizer">Randomize Enemies (Enemizer)</CheckBox>
 							<CheckBox UpdateToolTip="@UpdateToolTipID" Id="RandomizeFormationEnemizer" bind-Value="@Flags.RandomizeFormationEnemizer">Randomize Formations (Enemizer)</CheckBox>
+							<CheckBox UpdateToolTip="@UpdateToolTipID" Id="RandomizeEnemizer" IsEnabled="@Flags.RandomizeFormationEnemizer" bind-Value="@Flags.RandomizeEnemizer">Randomize Enemies (Enemizer)</CheckBox>
 						</div>
 						<div class="col-md-12">
 						</div>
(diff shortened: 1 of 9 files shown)


one1fineday:
The recent extra commit is there to prevent an issue where occasionally the same enemy would be selected twice in a formation.

one1fineday:
Fixed another error with the Zozo Melmond fix (now it checks if either Randomize Enemies OR Randomize Formations is on, before it was only checking for formation generation)

one1fineday:
Hopefully this fixes a potential issue with Domain placement not happening (this is the first time I've encountered a problem).  Originally I planned for Domain placement to be a separate feature but that is not going to happen, so now it will execute immediately after Formation Generation.

one1fineday:
Added a sanity check

one1fineday:
Fixed another stupid mistake I overlook.  Now boss fights should be properly unrunnable.
Note that unrunnable normal formations now will simply generate a random ambush chance of 3 to 30, without regard for surprise tiers, which is something to consider when shuffling unrunnables.

one1fineday:
I've put up a revamp of the code (again) which removes a lot of the clutter and disorganization in Enemizer.  I've also fixed a few bugs that have cropped up, and did some rebalancing changes regarding where monsters are placed.

one1fineday:
Added a check in flags.cs if Enemizer is turned on when pure and soft potions are required in Coneria shop.

Steven-Biro:
@one1fineday please let me know when this is ready to be merged, it looks like it is still under active development by you (updated4 hours ago)

one1fineday:
The current version on my github should be good to go.  If there are issues I haven't been able to discover them yet, and someone testing can probably find them better than me testing myself.

one1fineday:
Fixed a bug in the game's programming that caused the third monster in a 2-6 mixed formation to render with the wrong palette.

This should resolve all issues that I have seen so far.

FF1Lib/FF1Rom.cs
```diff
@@ -364,12 +365,12 @@ public void Randomize(Blob seed, Flags flags, Preferences preferences)
 				FiendShuffle(rng);
 			}
 
-			if (flags.FormationShuffleMode != FormationShuffleMode.None)
+			if (flags.FormationShuffleMode != FormationShuffleMode.None && !flags.RandomizeEnemizer && !flags.RandomizeFormationEnemizer)
```
latzio: Please make a bool function on flags that returns the result of ```!flags.RandomizeEnemizer && !flags.RandomizeFormationEnemizer```

Something like
`bool EnemizerEnabled => flags.RandomizeEnemizer || flags.RandomizeFormationEnemizer;`

and use it in all the places in this file where you need it.

FF1Lib/FF1Rom.cs
```diff
@@ -345,7 +346,7 @@ public void Randomize(Blob seed, Flags flags, Preferences preferences)
 			// Put this before other encounter / trap tile edits.
 			if ((bool)flags.AllowUnsafeMelmond)
 			{
-				EnableMelmondGhetto(flags.RandomizeFormationEnemizer);
+				EnableMelmondGhetto(flags.RandomizeFormationEnemizer || flags.RandomizeEnemizer);
```
latzio: Instead of passing this here - can we make the method that adjusts all the formations and separate it out. Something like ```void updateEncounterTiles() { ... }``` and just call it at the top of this method like we do with `ShiftEarthOrbDown()`

Then everything else after can assume the rewritten data and we don't need any checks like this.

FF1Lib/FF1Rom.cs
```diff
@@ -167,9 +167,10 @@ public void Randomize(Blob seed, Flags flags, Preferences preferences)
 			}
 #endif
 
-			if (flags.RandomizeFormationEnemizer)
+			if (flags.RandomizeEnemizer || flags.RandomizeFormationEnemizer)
 			{
-				DoEnemizer(rng, false, flags.RandomizeFormationEnemizer, false);
+				FixEnemyPalettes(); // fixes a big in the original game's programming that causes third enemy slot's palette to render incorrectly
```
latzio: `FixEnemyPalettes()` is totally harmless right? It's a pure fix? You can just call it at the top of Randomize() like with do with a bunch of other things.

FF1Blazorizer/Pages/Randomize.cshtml
```diff
@@ -463,7 +463,8 @@
 							<div class="checkbox-cell"></div>
 							<CheckBox UpdateToolTip="@UpdateToolTipID" Id="DisableTentSaving" bind-Value="@Flags.DisableTentSaving">Disable Tent/Cabin/House Saving on Overworld</CheckBox>
 							<CheckBox UpdateToolTip="@UpdateToolTipID" Id="DisableInnSaving" bind-Value="@Flags.DisableInnSaving">Disable Inn Saving</CheckBox>
-							<CheckBox UpdateToolTip="@UpdateToolTipID" Id="RandomizeFormationEnemizer" bind-Value="@Flags.RandomizeFormationEnemizer">Randomizer Formations (Enemizer)</CheckBox>
+							<CheckBox UpdateToolTip="@UpdateToolTipID" Id="RandomizeEnemizer" bind-Value="@Flags.RandomizeEnemizer">Randomize Enemies (Enemizer)</CheckBox>
```
latzio: You can add sections like ` IsEnabled="@Flags.RandomizeEnemizer"` to all the checkboxes that are incompatible with the Enemizer. In other places you check and disable them but you can do it in the UI too so users understand why things they've checked off (like formation shuffle) won't work.

FF1Lib/FF1Rom.cs
latzio: If that's not clear, I mean the function that changes all the `0x0A80` to `0x0A00` everywhere so that `0x0A00` is now the only way random encounters will work.

one1fineday: I suggested that and nitz said he did it.  I could write such a function and update everything that uses it.  However I still need to pass the flag to MelmondGhetto because I write my own formations rather than the hardcoded ones the original function uses.

FF1Blazorizer/Pages/Randomize.cshtml
one1fineday: Okay, I got that done.  I've also made it so that randomizing enemies requires selecting Generate New Formations first.  I've also renamed the options "Generate New Formations" and "Generate New Enemies" so that it is clear that this is not a standard shuffle.

The fun% flags will not be turned on/off since players store that information between seed generation, and I don't want to mess with that.

FF1Lib/FF1Rom.cs
one1fineday: Yes, it just removes an LSR A in the middle of the game's programming.  The disassembly notes it as a bug.

one1fineday:
Okay, this changes Flags.cs, FF1Rom.cs, and Randomize.cshtml to do what you asked, with the caveats I mentioned in comments above.  If you like, I can write the function to make translating map tiles to the new format as a general step the randomizer takes, because by now I am well acquainted with every possible place the randomizer could mess with that.  I would still keep trap tile shuffle disabled with Formation Generator though for a number of reasons (for one, the trap tiles are already effectively "shuffled" with the exception of Phantom which for various reasons I intended as a boss monster), and Formation Shuffle would be entirely superfluous with formation generator anyway.

Based on what I have seen, Warmech Unleashed should work without any issues.

one1fineday:
This latest commit adds a few things:
- WarMech encounter's ambush rate has been fixed (it's 75 instead of 4, which is what it should be)
- Tier 5 enemies which are assigned a script use WzVAMP's script instead of MudGOL's.
- I included a function which calculates the spell tier of a given spell (from the perspective of a regular enemy), which will be useful for a feature I intend to build much later.  The Enemizer now runs this function to calculate the spell tier instead of loading from a predefined list.

PR #294, opened 2019-08-07
Flags fix again

one1fineday:
Should fix the issue with flagstrings not loading the correct checkboxes.


one1fineday:
I fixed a bug in Spellcrafter where occasionally instant death spells did not have a correct branching path and thus left the spell blank.  This also means that you'll actually see time-elemental death spells now.

one1fineday:
Fixed a bug where LOCK spells' placement code was not working properly (LOCK could appear at the same level twice, rather than being spaced a minimum of three levels apart as intended).

FF1Lib/Spellcrafter.cs
```diff
@@ -316,7 +316,7 @@ enum OOBSpellEffects
 							validroutines.Add(0x0E);
 						else
 						{
-							if (SpellTier(lockspell) - SpellTier(index) > -2 || SpellTier(lockspell) - SpellTier(index) < 2)
+							if (SpellTier(lockspell) - SpellTier(index) > 2 || SpellTier(lockspell) - SpellTier(index) < -2)
```
latzio: If you use Math.abs( ) you don't have to compare to both positive and negative 2 and you will avoid errors like this. 

```
If (Math.abs(SpellTier(lockspell) - SpellTier(index) > 2)) ...
```

You have a few examples in this file you can update.

one1fineday:
Done, pushed the revised code.  It should work without breaking, I hope.

one1fineday:
Latest version uses Math.Abs to clean up the code.
I also allowed the Spellcrafter to create non-elemental AoE spells between from tier 3 to tier 5, and boosted the strength of non-elemental and time-elemental damage spells to make them more competitive with elementals.  Non- or Time- elementals are obviously great to have if they're available, but Red Mages can't use them and few enemies are actually weak to them.  They always have much better accuracy than the elemental spells though, since non-elementals at high tiers will have max accuracy, something elementals can never have.

PR #303, opened 2019-10-20
Numerous minor features (Enemizer and Spellcrafter)

one1fineday:
New Feature: Knight and Ninja gain charges in all levels
When checked, if Knight or Ninja gain charges at any level, they gain a charge in ALL levels.  This only applies to promoted characters and will never exceed their normal cap.  This gets around the issue where Knights and Ninjas had no access to spells they could learn due to lack of charges, and has no effect if Keep Permissions or Magic Level Shuffle is off,

New Feature: Free Tail
Does exactly what you think it would do, gives the player a free Tail at game start.  So long as they can find Bahamut, they are guaranteed promotion.  This does not prevent the player from finding another Tail in a treasure chest.

Balance Tweak: When Enemy Skill/Spell shuffle is on, Enemizer does not assign scripts to enemies which are normally guaranteed a skill list, with the exception of Magic Dragons and Magic Elementals/Djinns.  This could have led to seeds where many enemies would have NUKE or some other powerful spell, and is a safety precaution that would apply to most uses of Script Shuffle (unless Skill/Spell shuffle were turned off).


FF1Lib/Hacks.cs
```diff
@@ -537,6 +537,11 @@ public void EnableFreeLute()
 			Data[0x3020 + (int)Item.Lute] = 0x01;
 		}
 
+		public void EnableFreeTail()
+		{
+			Data[0x3020 + (int)Item.Tail] = 0x01;
+		}
+
```
latzio: This isn't quite good enough - and finishing it off will be easy. Look for like, "FreeAirship" elsewhere in the code and flesh it out so that if the TAIL is free it's replaced with a cabin, and no longer incentivized, and that state is correctly represented in the UI.

I think you'll just need to edit IncentivesData.cs and ItemPlacement.cs

The rest of this patch is pretty great through.

one1fineday:
How do I do that?  I haven't messed with the placement code but I tried tracing the path of the code when Treasure shuffle is on.  Is it enough to add FreeTail to the IVictoryConditionsFlags interface and replace as it does with the Lute and other free items?

one1fineday:
Added an invisible flag "IncentivizePromotion" which is distinct from IncentivizeTail.  If this is confusing I can invert the name so that IncentivizeTail is the final confirmation and IncentivizePromotion is the name of the button.

If my understanding of the placement code is correct, this should correctly replace the Tail with a Cabin.  I have not tested this yet.

one1fineday:
Spellcrafter Balance Tweak: Tier 8 Evasion and Armor always have the strength of self-cast for that tier, but target all allies.  Tier 8 attack up uses the formula for single-target but is multitarget.  For reference this means that:
INV8 = +200 evasion for all party members
FOG8 = +54 absorb for all party members
TMP8 = +28 attack for all party members
The formulae for FOG/GARD and TMPR/SABR have also been altered to nerf them slightly.

Tier 2 damage spells now have a chance to be multitarget, but only if they are attacks of the following elements: Death, Poison, Time, Status.  The accuracy of these spells is 24, like other AoE (normal single target spells are accuracy 40).  The effectivity of the spell is 70% of its single-target damage.

one1fineday:
Fixed error in comments for Spellcrafter (INVS spells add +20 at higher tiers, meaning INV8 offers +240 evade instead of +200)

Spider enemies now are guaranteed to draw a random attack ailment except Stone, Death, or Confusion touch.  Their attacks are always status-elemental.  Spiders are also barred from touch attack perks for obvious reasons, which increases the likelihood that they will draw stat increases, resists, or +1 hit.  Spiders also now correctly have the possibility of drawing a weakness.

Gargoyles attack tier is reduced to Below Average (3).

Gators are now locked to 2 hits instead of drawing 2-3 hits.

FF1Lib/Hacks.cs
one1fineday: I put in the changes outlined below for "FreePromotion".  It works properly on the web page when counting key items/incentives, and as far as I know it should mean the tail is properly incentivized.

Also added some tweaks for balance / quality of life.

one1fineday:
Fixed an error where tier 2 AoEs were not spawning and tier 3 AoEs were weaker than normal.

one1fineday:
Balance change: Red Mages can now cast QUIK (or any self-casting haste spell), but they cannot cast FAST.  This should make RM an interesting melee option even early in the game, if QUIK is available.  SABR is still gated behind class change however.

one1fineday:
Evasion and Absorb increasing spells should not duplicate in the same tier (self-cast and single/multi-cast can appear in the same level though).  Sleep and Lock spells should not duplicate in the same tier.

I added a shuffle of the spell indices before the main spell generation loop runs.  This should mean that there is no implied favoritism towards spells appearing in particular slots.

one1fineday:
Enemizer update:
Enemies which draw a spell list now draw a random spell list based on their monster tier.  Most tiers will have 3 possible spellbooks, while the earliest tiers only have two and the highest two still only have one possible spell list (Evilman).  This should add more variety to forced caster enemies.
Fixed an issue where enemy names could repeat.  Now Enemizer correctly removes a prefix from the pool of possible monster names once it has been used for that monster type.

one1fineday:
Fixed an error where enemies would not be assigned a zone.  If the enemy's zonelist is empty after the assignment loop, it will run the assignment loop again except it will not regard skill tiers for zone placement.

one1fineday:
Removed some stuff that was written to Console in the Enemizer code, mostly related to bugtesting.
I updated the windows executable so that it may access the features which have been added to the randomizer since the Windows app was depreciated.  It is still very rough (presets do not work properly) but you should be able to generate a seed with the alpha flags now.

one1fineday:
Fixed an error where flag strings would not import properly on the website.

The Windows executable is now updated with the exact same functionality as the website, including Tri-State flags.  I have yet to test this thoroughly (as in, actually generate a seed).  Additionally, the format the executable asks for and creates when copying a seed is similar to the URL used for the website, except that the string generated by the executable leads with "localhost?s=" instead of the site address.  I imagine when this moves to an actual official release it will be changed to mention the website.

one1fineday:
Added Knight and Ninja gain spell charges in all levels and fixed a tooltip.

I believe this is ready to go.

one1fineday:
Well, my belief is yet again unfounded.  Fixed a bug with armor up spells at tier 8 not properly targeting all party members.

one1fineday:
Toned down the number of unrunnables Generate New Formations creates slightly (chance is rng(0, 37) + zone xp tier >= 40, or 8/38 for the ToFR encounters, where it was 8/30).  I may adjust this again in the future.

FF1Blazorizer/Pages/Randomize.cshtml
```diff
@@ -100,13 +100,13 @@
 							<TriStateCheckBox UpdateToolTip="@UpdateToolTipID" Indent IsEnabled="@Flags.EnemyFormationsUnrunnable" Id="everythingUnrunnableCheckBox" bind-Value="@Flags.EverythingUnrunnable">All Encounters are Unrunnable</TriStateCheckBox>
 							<TriStateCheckBox UpdateToolTip="@UpdateToolTipID" Id="unrunnablesStrikeFirstAndSurpriseCheckBox" bind-Value="@Flags.UnrunnablesStrikeFirstAndSurprise">Unrunnable First Strike &amp; Surprise</TriStateCheckBox>
 							<TriStateCheckBox UpdateToolTip="@UpdateToolTipID" Id="enemyFormationsSurpriseCheckBox" bind-Value="@Flags.EnemyFormationsSurprise">Enemy Surprise Bonus</TriStateCheckBox>
-							<EnumDropDown UpdateToolTip="@UpdateToolTipID" Id="formationShuffleModeDropDown" TItem="FormationShuffleMode" IsEnabled="@(!Flags.RandomizeFormationEnemizer)" bind-Value="@Flags.FormationShuffleMode">Enemy Formations:</EnumDropDown>
+							<EnumDropDown UpdateToolTip="@UpdateToolTipID" TItem="FormationShuffleMode" IsEnabled="@(!Flags.RandomizeFormationEnemizer)" bind-Value="@Flags.FormationShuffleMode">Enemy Formations:</EnumDropDown>
```
latzio: Why did you remove all the IDs in this file? Weren't they used for anything like tooltips?I don't even remember now.

one1fineday: i might have been using an older version of randomize.shtml.  i can get a newer version and redo the website page based on that.  it wasn't intended.

one1fineday: Anyway I restored the IDs / changed the IDs to what they are in the base repo.  The Diff should now show only the two new buttons I added.

one1fineday:
Reduced the top limit of XP in ToFR fights to 21000 (was 32000).

Fixed a bug where FOG spells could still duplicate in tier 8.

one1fineday:
Updated incentive logic:
If NPC Shuffle is completely off, Main Items are in the incentive pool, and Free Airship is turned on, the TNT will not be in the incentive pool.
If NPC Shuffle is completely off, Main Items are in the incentive pool, Free Airship is turned on, and Titan's Trove is not an incentive location itself, the Ruby will not be in the incentive pool.

I don't think this will break anything.

one1fineday:
Another update to incentive logic regarding Ruby and TNT.
If TNT or Ruby are guaranteed to be meaningless towards progression based on the flagset (does not lead to any treasures), it will not be included in the incentive pool if Other Quest Items are in the incentive pool.  This mostly applies when Free Airship or Free Canal are turned on and NPC Shuffle is off.

public bool? RequiredRuby => !EarlySage & !NPCItems & !FreeAirship;
public bool? UselessRuby => FreeAirship & !TitansTrove;
public bool? RequiredTnt => !NPCFetchItems & !NPCItems & !(FreeCanal | FreeAirship);
public bool? UselessTnt => !NPCFetchItems & FreeCanal;

one1fineday:
I found a bug where No Dan Mode was not working properly.  This has been fixed by telling NoDan to patch over the area in expanded ROM where the level up routine now resides.

I also fixed an error where I added attack-up spells to the caster item pool instead of evade-down spells as intended.

one1fineday:
Updated logic for useless TNT to de-incentivize the TNT if Fetch Quest Shuffle is off, Free Airship is on, and Open Progression is off (there would be no dock to Onrac, so your only access is via the airship and ship adds no accessibility you wouldn't already have).

Updated incentive logic to de-incentivize Ship if FreeAirship and !MapOpenProgression (ship adds no locations that wouldn't be accessible by airship, because the only access to Onrac is by airship.

one1fineday:
Fixed a bug where Red Mage/Wizard would gain access to HEL2 and HEL3 that wasn't supposed to be there.

one1fineday:
Enemizer and Spellcrafter settings are TriState now.
Added a feature "Retain Old Permissions" for Spellcrafter.  This replaces the old functionality where old permissions would be retained if Magic Level shuffle is on (since now the user can choose whether they want this on or not).

one1fineday:
Fixed a bug that caused crash when a NULL spell is drawn (this should hopefully happen only rarely, but now the NULL spell will appear).

PR #307, opened 2019-11-16
split out the weapon fixes flag

Steven-Biro:


FF1Blazorizer/Pages/Randomize.cshtml
```diff
@@ -436,6 +436,9 @@
 						</div>
 						<div class="col-md-6">
 							<h4>Balance</h4>
+							<CheckBox UpdateToolTip="@UpdateToolTipID" Id="doubleBBCritCheckBox" bind-Value="@Flags.BBCritRate">Double BB Crit Rate</CheckBox>
+							<CheckBox UpdateToolTip="@UpdateToolTipID" Id="doubleWeaponCritCheckBox" bind-Value="@Flags.WeaponCritRate">Double Weapon Crit Rate</CheckBox>
+							<CheckBox UpdateToolTip="@UpdateToolTipID" Id="increaseWepBonusesCheckBox" bind-Value="@Flags.WeaponBonuses">Increase Weapon Elemental and Type Bonuses</CheckBox>
```
latzio: I think we should promote the thief hit rate growth into a sibling of these since it makes the class so much more decent.

latzio: And it will legitimize it too.

Steven-Biro: Makes sense to me, I can see about doing that later this evening

PR #308, opened 2019-11-19
Splitting Unrunnables

one1fineday:
Here is the preliminary work I've done on splitting the unrunnables.  It has not been thoroughly tested.  At the moment it includes a test of the feature by changing the domain around Coneria to include formation 00 and formation 80 (00's b-side) and sets the b-side to unrunnable.  I believe the romhack portion works fine, but I have not thoroughly tested the changes I made to other randomizer features to ensure compatibility.

Additionally, ShuffleUnrunnables now checks the formations it cycles through with foreach, and if they were unrunnable normally (either with vanilla stats or with the formation generator's alterations), ShuffleUnrunnables will create that many unrunnables on the A-Side.  On the B-Side, all formations except the fiend fights are included in the shuffle.  This will mean that there are two extra potential unrunnables on the B-Side from the normally empty B-Sides of Vampire and Astos, but Enemizer uses those formation incides.

### DIFF
--- a/FF1Lib/Enemizer.cs
+++ b/FF1Lib/Enemizer.cs
@@ -740,7 +740,8 @@ namespace FF1Lib
 			public byte pal2;
 			public byte surprise;
 			public byte paletteassignment = 0b00000000;
-			public bool unrunnable;
+			public bool unrunnable_a;
+			public bool unrunnable_b;
 
 			public byte Top
 			{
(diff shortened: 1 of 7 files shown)


one1fineday:
Updated Enemizer unrunnability chances, and created a constant in Enemizer.cs to make editing the unrunnable chance easier.

one1fineday:
Fixed an oversight where Unrunnable Shuffle would include the fiend B-Sides in the shuffle, creating more unrunnables than there should be.

one1fineday:
Unrelated, but the threshold for a non-elemental damage spell being classified as "very strong" is decreased from 75 to 68.  This should mean Tiamat 2 isn't carrying Nuclear-level spells without skill shuffle.

one1fineday:
Removed the testing block which set the fights around Coneria, now that I know the ROM hack works.

one1fineday:
Unrelated change to Spellcrafter:
Red Wizards can now learn TMPR, LOCK, and AMAG spells after class change.
Red Mages can now learn GARD spells.

These changes should make Red Wizards a lot more viable for parties if class change can be attained.

one1fineday:
Spellcrafter no longer sets tier 8 elementals to 107 accuracy - the elementals are always accuracy 24 (multitarget) or 48 (single target)

PR #313, opened 2019-12-02
Only Bosses Script/Skill Shuffle Option and Change to SplitUnrunnables

one1fineday:
(Only Bosses Script/Skill Shuffle)
Adds an option to only shuffle the scripts and skills of the fiends (each is separate).  This could be changed to a button to explicitly include the fiends instead, but for now it just skips the normal enemies if onlybosses is checked.

(Change to Hacks.cs)
Now it moves the block of code in PrepareEnemyFormation forward, so that the padded code can be used for other functions of interest.

This has been tested and it doesn't break the game, so we should be good to go.


one1fineday:
Update:

"Double BB Crit Rate" checkbox is changed to "Halve BB Crit Rate", and in all presets it defaults to on.  Additionally, the randomizer now checks if this box is on to make changes to the ROM, instead of checking if the box is OFF.

This is in line with the user's expectation that checking a box in randomizer means something is being changed from the original game, because halving the BB crit rate is obviously not the game's default behavior.

one1fineday:
Updated the randomizer page so that tooltips will appear for the price/stat sliders and Free Tail.

I saw nothing really interesting in splitting the first fiends and final bosses, and didn't want to annoy the player with too many options if they just wanted the traditional shuffle.

PR #317, opened 2019-12-27
Many tooltip updates and text clarifications to randomizer page

aaronringgenberg:
* Fixed typos, moved or edited tooltips that had been copy-pasted but not changed
* Removed clearly redundant tooltips (e.g. Ribbon: Ribbon)
* Added clarity to numerous tooltips/descriptions
* Added tooltips where none existed
* Clarified key item XP scaling


aaronringgenberg:
So I'm closing this because apparently this is a deprecated file. Perhaps a deletion is in order by someone with more context.

PR #318, opened 2019-12-31
Minor tooltip update + Enemizer/Spellcrafter minor update

one1fineday:
Updates tooltip for "Halve BB Crit Rate" to correctly describe the new functionality.

### DIFF
--- a/FF1Blazorizer/wwwroot/tooltips/tooltips.json
+++ b/FF1Blazorizer/wwwroot/tooltips/tooltips.json
@@ -655,7 +655,7 @@
 	},
 	{
 		"Id": "doubleBBCritCheckBox",
-		"title": "Double BB Crit Rate",
+		"title": "Halve BB Crit Rate",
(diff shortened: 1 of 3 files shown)


FF1Blazorizer/wwwroot/tooltips/tooltips.json
```diff
@@ -656,7 +656,7 @@
 	{
 		"Id": "doubleBBCritCheckBox",
 		"title": "Double BB Crit Rate",
```
latzio: This should be Halve too right?

one1fineday: Right.  Fixed.

one1fineday:
Made a small modification to Enemizer and Spellcrafter.  If the game encounters a spell of tier 0 in an enemy script, it will insert a tier 1 spell instead.

Additionally, the spiked tiles in Temple of Fiends and Ordeals should not contain duplicates of the Wizard, Eye and ZombieD fights any more, as those fights were always the first rolled fights of formation 2 and 6.  The tiles now point to other formations in those tiers.

PR #334, opened 2020-04-09
Add smb message

blackknight36:
This is just for fun.  :)


FF1Lib/Credits.cs
```diff
@@ -9,6 +9,15 @@ public partial class FF1Rom : NesRom
...
 		private static readonly List<string[]> VictoryMessages = new List<string[]>
 		{
+			new [] // SMB
+			{
+				"",
+				"THANK YOU MARIO!",
+				"",
+				"BUT OUR PRINCESS",
+				"IS IN ANOTHER",
+				"CASTLE!",
```
latzio, line 19: In keeping with what we do, this should be more like 
THANK YOU LIGHT WARRIORS
BUT OUR PRINCESS IS IN
ANOTHER TIMELINE

blackknight36: I thought about that as well.  Message has been updated and should now look like this.

image

blackknight36:
Added a few more ending messages.  Use them if you want.

PR #356, opened 2020-05-08
Mystery Seeds

jtoyoda:
This PR adds a new way to play the randomizer: Mystery seeds!  This option is now available on the FE site. The player is only told how many loose items there are, which items are incentive and what locations are incentivized. It is up to them to determine everything else. 

image

This PR also includes a simple command line interface for testing which takes in a weights file and output the needed information to generate a seed. In the future more weight presets could be added to the website in the same way the flagset presets are created.

### DIFF
--- a/FF1Lib/MysterySeedGeneratorLib.cs
+++ b/FF1Lib/MysterySeedGeneratorLib.cs
@@ -1,9 +1,8 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.Linq;
-using FF1Lib;
 
-namespace MysterySeedGenerator
+namespace FF1Lib
 {
 	public static class MysterySeedGeneratorLib
 	{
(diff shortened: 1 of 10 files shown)


one1fineday:
I want to give this a closer look to see if everything will work.  It will take a few days to find the time.

FF1Lib/MysterySeedGeneratorLib.cs
```diff
@@ -0,0 +1,1280 @@
...
+					looseCount = 2;
+					break;
+				case LooseItemCounts.Loose3:
+					looseCount = 3;
+					break;
+				default:
+					throw new ArgumentOutOfRangeException();
+			}
+
+			// Determine incentive count and location count
```
jtoyoda, line 290: This is definitely the part that is the most likely to contain bugs (the whole incentive selection part). I tried to test it but it may require unit tests to fully validate.

There is one known issue which is the following:
- You need 10 loose and you have a total of 11 possible items including the ship and canal incentive.
- You select all 9 of the single loose items without selecting the ship and canal.
- It is now "impossible" to get up to 10.

This is not possible with the current weights (also extrememly unlikely in general) but I couldn't think of a good way to fix it while still maintaining the weights

jtoyoda:
This will need to be rebased against dev

PR #370, opened 2020-05-23
Multiple Changes

one1fineday:
I have a list of changes ready for testing.  They still need to be checked for potential errors, so I'm going to wait a bit on adding them to the alpha, but some more eyes would be helpful.  These changes do a few things:

- Save Game When Game Over: When the party wipes, the game is saved (with necessary conditions to prevent players from accidentally softlocking or cheating story battles like Garland and Astos).  NPCs which trigger a fight will always have a condition unrelated to the battle that determines whether they will be reset to visible.  If the player has the airship, they will respawn where they parked the airship and the ship will be moved to Coneria.  If the player has the ship but not the airship, they will respawn where they last docked the ship.  If the player has neither, they will respawn where they last saved.  In any event, the save is written after the player presses a button to clear the "party perished" checkbox, so if a player does not want to respawn by these rules (but lose all of their progress) they can manually reset the console.

- New Talk_Astos routine: For all seeds, Astos will behave similarly to Bikke, in that he only gives you his item AFTER you defeat him, rather than before.  Astos will say his normal speech before you fight him, and will either print the standard NPC shuffle gift-item text after you beat him or simply print "Hurray!!" if NPC shuffle is off (the jingle will only play when you actually receive the item).  I decided to make this an "always on" feature to minimize the amount of flag-checking that needs to be done with other features which affect the talk routines, like the planned Astos Shuffle.

- Enemizer Formation Shuffle will now shuffle the pattern tables so that enemies that couldn't appear together in vanilla can appear together in Enemizer.  To encourage more enemies to share formations (due to Enemizer's placement rules regarding palettes), Formation Shuffle will also assign palettes just like Enemy Generation does, though the underlying enemy does not change.

- Reinstated jtoyoda's "Generate ROM" patch, which I probably borked.  After this is put into the rando, I'm going to re-download everything and will probably have to mess with Visual Studio again.  Le sigh.

### DIFF
--- a/FF1Blazorizer/Pages/Randomize.razor
+++ b/FF1Blazorizer/Pages/Randomize.razor
@@ -65,7 +65,7 @@
         
 
         <br />
-        <button type="button" class="nes-btn is-primary" @onclick="@OnRandomize">Randomize</button>
+        <button type="button" class="nes-btn is-primary" @onclick="@OnRandomize">Generate ROM</button>
         <BSCollapse IsOpen="@(StatusMessage.Length > 0)">
             <div class="is-dark nes-container">
                 @StatusMessage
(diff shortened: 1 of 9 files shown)


one1fineday:
Updated it so Talk_Astos works as it is supposed to.
I've also re-introduced "'tis I, Astos, all along!!" as Astos' text in NPC shuffle (he doesn't specify which item he is holding).  I am going to test non-NPC shuffle tomorrow and we will see if the hack works properly there too.  If so (and if Enemizer works out which it should) then this is going in to alpha.

PR #380, opened 2020-06-02
Pool Party + Random Class Change

wildham0:
**Pool Party**
Change party creation at the start of the game. It fills a pool with random characters, allowing duplicates, from which to form the party. Once a character is selected, it's removed from the pool. Cancelling your choice puts back the character in the pool.
Key points:
 - Modify extensively DoPartyGen_OnCharacter
 - Use an array of 10 bytes in memory which isn't used at party creation, tests didn't bring up any problems
 - Ignore Party Composition flags for now, but it's possible to take them into accounts down the road. Works fine with Tavern Recruitment.

**Random Class Change**
Promote party to random classes.
Key points:
- Modify extensively DoClassChange, doesn't break "Nones"

Flags are in experimental section

### DIFF
--- a/FF1Blazorizer/Pages/Randomize.razor
+++ b/FF1Blazorizer/Pages/Randomize.razor
@@ -514,10 +514,10 @@
                         <TriStateCheckBox UpdateToolTip="@UpdateToolTipID" Id="RandomArmorBonus" @bind-Value="Flags.RandomArmorBonus">Random Armor Bonus</TriStateCheckBox>
                         <CheckBox UpdateToolTip="@UpdateToolTipID" Id="fixMissingBattleRngEntry" @bind-Checked="Flags.FixMissingBattleRngEntry">Fix Missing Battle RNG Entry</CheckBox>
                         <TriStateCheckBox UpdateToolTip="@UpdateToolTipID" Id="swolePiratesCheckBox" @bind-Value="Flags.SwolePirates">Buffed Pirates</TriStateCheckBox>
-                        <CheckBox UpdateToolTip="@UpdateToolTipID" Id="EnablePoolParty" @bind-Checked="Flags.EnablePoolParty">Select Party From Pool of Random Characters</CheckBox>
+                        <TriStateCheckBox UpdateToolTip="@UpdateToolTipID" Id="EnablePoolParty" @bind-Value="Flags.EnablePoolParty">Select Party From Pool of Random Characters</TriStateCheckBox>
                         <EnumDropDown UpdateToolTip="@UpdateToolTipID" Indent IsEnabled="@Flags.EnablePoolParty" Id="PoolSizeDropDown" TItem="PoolSize" @bind-Value="Flags.PoolSize">Pool Size :</EnumDropDown>
                         <TriStateCheckBox UpdateToolTip="@UpdateToolTipID" Indent IsEnabled="@Flags.EnablePoolParty" Id="PPIncludePromotedClasses" @bind-Value="Flags.IncludePromClasses">Include Promoted Classes</TriStateCheckBox>
-                        <CheckBox UpdateToolTip="@UpdateToolTipID" Id="EnableRandomPromotions" @bind-Checked="Flags.EnableRandomPromotions">Random Classes at Class Change</CheckBox>
+                        <TriStateCheckBox UpdateToolTip="@UpdateToolTipID" Id="EnableRandomPromotions" @bind-Value="Flags.EnableRandomPromotions">Random Classes at Class Change</TriStateCheckBox>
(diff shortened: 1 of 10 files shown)


Steven-Biro:
@wildham0 from a quick skim, this looks really well put together, I have one request as of right now, a comment in the 1E.asm file about the changes to DoPartyGen_OnCharacter and pointing out the other asm files there so people in the future dont get mixed up trying to figure out what that routine is supposed to do

wildham0:
thanks, I'll add the comments in the asm files

FF1Blazorizer/Pages/Randomize.razor
```diff
@@ -514,6 +514,11 @@
                         <TriStateCheckBox UpdateToolTip="@UpdateToolTipID" Id="RandomArmorBonus" @bind-Value="Flags.RandomArmorBonus">Random Armor Bonus</TriStateCheckBox>
                         <CheckBox UpdateToolTip="@UpdateToolTipID" Id="fixMissingBattleRngEntry" @bind-Checked="Flags.FixMissingBattleRngEntry">Fix Missing Battle RNG Entry</CheckBox>
                         <TriStateCheckBox UpdateToolTip="@UpdateToolTipID" Id="swolePiratesCheckBox" @bind-Value="Flags.SwolePirates">Buffed Pirates</TriStateCheckBox>
+                        <CheckBox UpdateToolTip="@UpdateToolTipID" Id="EnablePoolParty" @bind-Checked="Flags.EnablePoolParty">Select Party From Pool of Random Characters</CheckBox>
```
Steven-Biro, line 517: is there a reason some of these are not tri state and other are?

it seems like there isnt a reason to not have some of these as tri-state

FF1Blazorizer/wwwroot/tooltips/tooltips.json
```diff
@@ -94,6 +94,31 @@
 		"screenshot": "allowUnsafePiratesCheckBox.png",
 		"description": "Unsafe Pirates: Allows Pirates to be shuffled skills and spells. Left unchecked, Pirates will always have a script that directs them to only attack."
 	},
+	{
```
Steven-Biro, line 97: thank you for including tooltips!

FF1Lib/FF1Rom.cs
```diff
@@ -963,10 +975,10 @@ public void UpgradeToMMC3()
 		public void WriteSeedAndFlags(string version, string seed, string flags)
 		{
 			// Replace most of the old copyright string printing with a JSR to a LongJump
-			Put(0x38486, Blob.FromHex("20FCFE60"));
+			Put(0x38486, Blob.FromHex("20B9FF60"));
 
 			// DrawSeedAndFlags LongJump
-			PutInBank(0x1F, 0xFEFC, CreateLongJumpTableEntry(0x0F, 0x8980));
+			PutInBank(0x1F, 0xFFB9, CreateLongJumpTableEntry(0x0F, 0x8980));
```
Steven-Biro, line 981: is there a reason this was moved? I didnt believe there was much room left for moving these longjumps around

-- I really need to get around to fixing that

FF1Lib/Flags.cs
```diff
@@ -730,6 +743,11 @@ public static Flags DecodeFlagsText(string text)
 				ClampBossHPScaling = GetTriState(ref sum),
 				SeparateEnemyHPScaling = GetTriState(ref sum),
 				SeparateBossHPScaling = GetTriState(ref sum),
+				PoolSize = (PoolSize)GetNumeric(ref sum, Enum.GetValues(typeof(PoolSize)).Cast<int>().Max() + 1),
+				EnablePoolParty = GetBoolean(ref sum),
+				IncludePromClasses = GetTriState(ref sum),
+				EnableRandomPromotions = GetBoolean(ref sum),
+				IncludeBaseClasses = GetTriState(ref sum),
```
Steven-Biro, line 750: These need to be in the same place in the encoding and decoding order, notice how the spoilers flag is at the end of the encoding and its at the end of the decoding

FF1Blazorizer/Pages/Randomize.razor
wildham0: I'm just thinking that you wouldn't want to tri-state a specific feature, but you would for its options, but that might not be how other use tri-state.

FF1Lib/FF1Rom.cs
wildham0: CHRload isn't very longjump friendly, and there's just enough free space there to put the new routine without the DrawSeedAndFlags long jump, so I moved it to a smaller corner of free space.

FF1Blazorizer/Pages/Randomize.razor
Steven-Biro: Some people like torturing themselves (also sometimes things like this are used in the challenge weeklies so people dont know going in, but find out right away when they start the game if something is on or not)

wildham0:
I think I don't have the necessary access to merge.

Steven-Biro:
oh, I thought approving it would allow you to, ok well I'll merge then

wildham0:
thanks!

PR #384, opened 2020-06-07
ShuffleAstos and SaveWhenGameOver fix

wildham0:
SaveWhenGameOver fixes
Fix a bug with SaveWhenGameOver that would revives Nones when reloading a game; also change the main routine so has to not fiddle with NPC flags as much.
Key points:
- Change from checking various NPC flags to checking which battle killed the party.
- Restore Astos to normal behaviour without allowing to cheat his item.

SuffleAstos
Shuffle Astos script to a random key NPC, as is he's hiding as this new NPC.
Key points:
- Compatible with SaveWhenGameOver
- If Astos is Bahamut, there's substantial change to Astos's routine and SaveWhenGameOver

### DIFF
--- a/FF1Lib/Flags.cs
+++ b/FF1Lib/Flags.cs
@@ -686,8 +686,8 @@ namespace FF1Lib
 			sum = AddBoolean(sum, flags.FiendShuffle);
 			sum = AddBoolean(sum, flags.DisableTentSaving);
 			sum = AddBoolean(sum, flags.DisableInnSaving);
-			sum = AddBoolean(sum, flags.SaveGameWhenGameOver);
 			sum = AddBoolean(sum, flags.ShuffleAstos);
+			sum = AddBoolean(sum, flags.SaveGameWhenGameOver);
 			sum = AddTriState(sum, flags.RandomizeEnemizer);
 			sum = AddTriState(sum, flags.RandomizeFormationEnemizer);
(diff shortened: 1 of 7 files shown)


FF1Lib/Flags.cs
```diff
@@ -251,6 +252,8 @@ public class Flags : IIncentiveFlags, IMapEditFlags, IScaleFlags, IFloorShuffleF
 		public bool? ClampEnemyHpScaling { get; set; } = false;
 		public double EnemyHPScaleFactor { get; set; } = 1;
 		public double BossHPScaleFactor { get; set; } = 1;
+		public bool EnablePoolParty { get; set; } = false;
+		public bool EnableRandomPromotions { get; set; } = false;
```
Steven-Biro, line 256: part of your other PR included by accident?

Steven-Biro: noticed that this assumes your other PR is already included, thats fine, this should all resolve once that one is merged in

So you know for the future, you can target your other branch with this PR and when that one is merged in, this one will change to be a PR against alpha


alpha <-- first PR <-- second PR

merge PR 1 and it automatically becomes

alpha <-- second PR

this is pretty recent and its really nice for things like this.

Steven-Biro:
@wildham0 theres a conflict now if you want to fix that, I merge it directly after

wildham0:
fixed

PR #389, opened 2020-06-17
use the seed and flags for rng

Steven-Biro:
I think I'm probably over complicating this, but it should work.

### DIFF
--- a/FF1Lib/FF1Rom.cs
+++ b/FF1Lib/FF1Rom.cs
@@ -84,12 +84,9 @@ namespace FF1Lib
 			MT19337 rng;
 			using (SHA256 hasher = SHA256.Create())
 			{
-				string flagstring = Flags.EncodeFlagsText(flags);
-				byte[] FlagsBytes = hasher.ComputeHash(Encoding.UTF8.GetBytes(flagstring));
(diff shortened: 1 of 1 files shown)


FF1Lib/FF1Rom.cs
```diff
@@ -81,7 +81,17 @@ public static async Task<FF1Rom> CreateAsync(Stream readStream)
...
-			var rng = new MT19337(BitConverter.ToUInt32(seed, 0));
+			MT19337 rng;
+			using (SHA256 hasher = SHA256.Create())
+			{
+				string flagstring = Flags.EncodeFlagsText(flags);
+				byte[] FlagsBytes = hasher.ComputeHash(Encoding.UTF8.GetBytes(flagstring));
+				byte[] SeedAndFlags = new byte[seed.Length + FlagsBytes.Length];
+				Array.Copy(seed, SeedAndFlags, seed.Length);
+				Array.Copy(FlagsBytes, 0, SeedAndFlags, seed.Length, FlagsBytes.Length);
+				byte[] hash = hasher.ComputeHash(SeedAndFlags);
```
Entroper, line 92: I think you can do something like:

    flagsBytes = Blob.FromBase64(flagstring);
    hash = hasher.ComputeHash(flagsBytes + seed);

PR #413, opened 2020-08-06
Alpha

iwritecode791:
Made the tri-state clamps reflect what the two options are a bit better.

Fixed an issue with the lower range going above the higher and saving incorrectly in the flag string.


iwritecode791:
Didn't add description...
Bug fix for low going above high causing the flag string to save the values incorrectly.

PR #440, opened 2020-09-01
Fix all the Warnings.

latzio:
The only non obvious one is avoiding the unreachable code by changing
the const variables to static readonly. Then C# feels differently about
those comparisons.

The Sandbox is still useless but it builds now.

I fixed a bug dragging around the tooltips so it's not super annoying now and actually captures mouse events properly!

### DIFF
--- a/FF1Lib/FFRVersion.cs
+++ b/FF1Lib/FFRVersion.cs
@@ -3,9 +3,9 @@
 	// THIS FILE IS UPDATED DURING THE BUILD AND DEPLOY. DO NOT EDIT UNLESS YOU KONW WHAT YOU ARE DOING.
 	public static class FFRVersion
 	{
-		public const string Version = "3.1.1";
-		public const string Sha = "SHA";
-		public const string Branch = "BRANCH";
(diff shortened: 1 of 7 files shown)


FF1Lib/FFRVersion.cs
```diff
@@ -3,9 +3,9 @@
...
 	{
-		public const string Version = "3.1.1";
-		public const string Sha = "SHA";
-		public const string Branch = "BRANCH";
-		public const string MasterBranch = "master";
+		public static readonly string Version = "3.1.1";
+		public static readonly string Sha = "SHA";
+		public static readonly string Branch = "BRANCH";
+		public static readonly string MasterBranch = "master";
 	}
```
Entroper, line 10: What were the warnings here?

latzio: The warnings were actually about unreachable code doing an if check on what the value was because it was constant and thus "always false" 

Of course we have a script that edits these....

Entroper: Hah, cool.

PR #444, opened 2020-09-10
Randomize page fixes, dungeon npc hints fix, No tail flag, Class as NPC flags

wildham0:
* Fix Experimentals tab bug
* Fix Blursed Armor bonus slider
* Reinsert Party Comp in Party tab
* Fix Dungeon NPC hinters not working
* New flag: Remove Tail
* New flags: Classes as NPC


FF1Blazorizer/Pages/Randomize.razor
```diff
@@ -28,7 +28,7 @@
 		<PartyTab Flags="@Flags" UpdateToolTipID="@UpdateToolTipID" />
 		<ConveniencesTab Flags="@Flags" UpdateToolTipID="@UpdateToolTipID" />
 		<BugFixesTab Flags="@Flags" UpdateToolTipID="@UpdateToolTipID" />
-		<ExperimentalTab Flags="@Flags" UpdateToolTipID="@UpdateToolTipID" IsOpen="false" />
```
latzio, line 31: This was deliberate of course. Now that there is like 4 things we don't expect people to use I wanted to hide it by default.

wildham0: Oh yeah, but this way when you tried to click on a flag it in Experimental it kept closing it, it's still hidden by default with this change.

latzio: Oh that's crazy. Thanks!

PR #455, opened 2020-11-14
Versioned deploys

Steven-Biro:
This is up for feedback right now only, and to see what netlify does with this as a PR, it should just deploy the standard thing with nothing changing.

The intent of this change is to create new sites each time we deploy the main site.

So when www.finalfantasyrandomizer.com updates to version 5.23.1, we create 5-23-1.finalfantasyrandomizer.com and deploy the exact same build there. A possible additional change that I am thinking about including is the flag decoding, it should check the version only for the versioned and live site, but the sha for alpha and beta sites. This would allow for minor updates that do not affect flags or rng to each version after it is deployed.

### DIFF
--- a/.circleci/deploy.sh
+++ b/.circleci/deploy.sh
@@ -41,11 +41,11 @@ else
     --data-raw "{\"custom_domain\": \"${version}.finalfantasyrandomizer.com\", \"force_ssl\": \"true\"}")
     
     
-    errorsExist=$(echo "$deploy" | jq ".errors!=null")
-    if [ ! -z "$errorsExist" ]; then
-    	echo "errors encountered while creating site:"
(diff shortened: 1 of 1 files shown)


.circleci/deploy.sh
```diff
@@ -22,5 +23,34 @@ if "$deployPreview"; then
     curl -X POST -H "Accept: application/vnd.github.v3+json" -H "Content-Type:application/json" "$pr_comment_url" -u $GH_USER:"$GH_API" -d "$post_data"
 
 else
+    version=$(grep " Version.*" /root/ff1randomizer/FF1Lib/FFRVersion.cs | grep -Eo "[0-9\.]+" | tr '.' '-')
```
Entroper, line 26: If we do this by tag, we can just use `$CIRCLE_TAG`.

PR #459, opened 2020-11-17
Added randomized incentive locations within selected dungeons.

iwritecode791:
As title. The implications of this are that users can select which dungeon they want to have an incentive item show up but not know exactly which chest. This should improve race-ability without having to rely on knowing exactly which chest they need to open.


wildham0:
This looks sound to me, we can commit this to beta after the next live update.

latzio:
I does look like it would work. However it *could* be made to work with entrance and floor shuffle as well, which I would like to see done at some point.

iwritecode791:
It does have a fun and interesting style with floor/entrance as is: If you incentivize volcano, that means some volcano floor will contain an incentive item. You could then cross that off your list and be able to skip the other volcano floors.

I think you wanted it to be that the volcano entrance would have an incentive item, this is another idea that could be implemented, but I think the random incentive has enough merit with floor/entrance to not overwrite that behavior.

iwritecode791:
I made that fix that wildham and I discussed earlier about titan's trove. Its behavior will be in line with all other locations.

PR #469, opened 2021-01-03
Extensive Hint Flag

mhn65536:
The new Flags replace a lot of npc dialogs with hints.

Currently there a 6 Hint Categories:
- Loose Floor: provides the floor where a loose is located(if there is more than one loose on that floor multiple hints a generated)
- Loose Name: provides the name of the loose and the dungeon
- Incentive Name: provides the name of the incentive ant the dungeon/npc/town
- Equipment Floor: States on what floors legendery items are
- Equipment Name: Gives the name of the Equipment and the dungeon name
- Floor Hints:
- - Wow: Quest or legendary
-  - Meh: Rare Equipment
-  - Sparkle: more than 5kgold treasure
-  - Dud: none of the above

It can be configures what hintcategories may end up where and how many "percent" of a category will be placed.

The hint categories are placed in the order specified. If there are no valid places left, remaining hints are not placed anymore. That is the reason, why you can specify the processing order for each category.

The flag is meant to be played with a lot of loose items. By default alle of the "Loose Floor" hints will be present and in the inner sea. So it will be a little like you are told where the incentive chests are at the start of the game instead of fixed ones.

If someone has some better names for the categories etc. I'd be delighted.

I tested it so far only in coneria and coneria castle, because i'm completly incompetent at the game.

regards
MadMartin


FF1Lib/Flags.cs
```diff
@@ -898,8 +1021,30 @@ public static Flags DecodeFlagsText(string text)
 		{
 			var sum = StringToBigInteger(text);
 
+			var tempExtensiveHints_BinMatrix = new bool[Enum.GetValues(typeof(HintCategory)).Length][];
+
+			for (int i = 0; i < tempExtensiveHints_BinMatrix.Length; i++) tempExtensiveHints_BinMatrix[i] = new bool[Enum.GetValues(typeof(HintLocation)).Length];
+
+			//You've got to be fucking shitting me. I have to do it in reverse?
```
latzio, line 1028: Obviously the flags do need to be decoded in reverse of the order they are encoded since the whole thing is just one bigass number written down as a hex string. I think you can remove the comment.

latzio: Generally humorous comments and variable names are tolerated but this is probably a bridge too far.

FF1Lib/ItemLists.cs
```diff
@@ -138,6 +138,13 @@ public static class ItemLists
 				.Where(x => x >= Item.Gold10 && x <= Item.Gold65000)
 				.ToList();
 
+		public static readonly IReadOnlyCollection<Item> BigGoldTreasure =
+			Enum.GetValues(typeof(Item))
+				.Cast<Item>()
+				.Where(x => x >= Item.Gold5000 && x <= Item.Gold65000)
```
wildham0, line 144: Careful, when price scaling is enabled the values can wildly change

FF1Lib/ExtensiveHints/Translator.cs
```diff
@@ -0,0 +1,240 @@
...
+				{MapLocation.IceCave5,"IceCave 5"},
+				{MapLocation.IceCaveBackExit,"IceCave 6"},
+				{MapLocation.IceCaveFloater,"IceCave 4"},
+				{MapLocation.IceCavePitRoom,"IceCave 4"},
+				{MapLocation.SeaShrine1, "SeaShrine 1"},
+				{MapLocation.SeaShrine2, "SeaShrine Right 2"},
+				{MapLocation.SeaShrine2Room2, "SeaShrine Right 2"},
+				{MapLocation.SeaShrine4, "SeaShrine LEft 2"},
+				{MapLocation.SeaShrine5, "SeaShrine Left 3"},
+				{MapLocation.SeaShrine6, "SeaShrine LEft 4"},
```
wildham0, line 93: Typos

FF1Blazorizer/Pages/Randomize.razor
```diff
@@ -19,6 +19,7 @@
 	<BugFixesTab Flags="@Flags" UpdateToolTipID="@UpdateToolTipID" />
 	<ExperimentalTab Flags="@Flags" UpdateToolTipID="@UpdateToolTipID" />
 	<FunTab Flags="@Flags" UpdateToolTipID="@UpdateToolTipID" />
+	<HintsTab Flags="@Flags" UpdateToolTipID="@UpdateToolTipID" />
```
wildham0, line 22: It should probably be put before the Conveniences Tab

FF1Lib/Flags.cs
```diff
@@ -547,6 +547,101 @@ public class Flags : IIncentiveFlags, IMapEditFlags, IScaleFlags, IFloorShuffleF
 			+ ((IncentivizeTitansTrove ?? true) ? 1 : 0)
 			+ ((IncentivizeSkyPalace ?? true) ? 1 : 0);
 
+		#region ExtensiveHints
+
+		public bool ExtensiveHints_Enable { get; set; } = true;
```
wildham0, line 552: Don't enable flags by default, this can mess up with presets

wildham0 (changes requested):

FF1Lib/Flags.cs
mhn65536: Sorry.

FF1Lib/ItemLists.cs
mhn65536: Question beeing, if that is a problem. I took 5000 because it was somewhat in the middle an a nice number. What would you propose?

FF1Lib/Flags.cs
mhn65536: done, was for testing...

FF1Blazorizer/Pages/Randomize.razor
mhn65536: done

FF1Lib/ExtensiveHints/Translator.cs
mhn65536: done

FF1Lib/Flags.cs
mhn65536: removed

mhn65536: When serializing and deserializing things, you usually do it in the same order. While i understand the implementation and the reason for it, it is very unintuitive. Contributors should not have to learn the quirks of a specific implementation. Therefore i come tho think, that there actually should be a comment. But that is not the scope of this PR.

mhn65536:
I should now be in a testable state again.

FF1Lib/ItemLists.cs
wildham0: What I mean is the item Gold5000 can turn out to be 2500 GP or 10,000 GP with price scaling 50-200, and the item Gold3400 can end up more valuable than Gold5000, basically the name don't represent the real value, it's something to be aware of

mhn65536: i see, to be corrected then, is there an easy way to check the actual values?

wildham0: The final value is the price, so you can probably cross-reference it (see ScalePrices in Scale.cs) :
		public const int PriceOffset = 0x37C00;
		public const int PriceSize = 2;
		public const int PriceCount = 240;

mhn65536: ok

mhn65536: should have been 10000g in the commit message

PR #470, opened 2021-01-11
Starting Inventory/Shop Reduction

mhn65536:
### Starting Inventory
Under the Tab scale you can configure your Starting Inventory(Tents, Cabins...)

### Shop Reduction
Reduces the number of Items/Magic sold in Shops, or removes Shops entirely. You can configure the Mode and Amount of Reduction for each shop type. The Coneria shops can be excluded from the reduction.

Modes:
- None, do not remove anything
- Whole Shops, removes the specified amount of shops as a whole
- AllItemsOfAKind, selects the specified amount of distinct Items and removes all instances of them(eg. all Heals)
- Spread, applies the reduction for each shop individually(eg. each shop gets a 40% reduction)
- Unequal, removes the amount from all entries(eg. some shops will be full, some might be half full, some might be removed)

Quest Items will never be removed.
Caravan Shop will not be removed from map. If it rolls empty, it just sells softs.

ShopShuffle, Inn/Clinic Prices reworked for common data access into the rom.
ShopShuffle algorithm altered, to always fulfill EarlyAilments and NoDuplicates constraints first try.

Sorry, changed some Line Endings...


wildham0:
SartingInventory looks sound, altho I'm going to question the level of granularity offered: there's mostly no differences between 1 or 2 items except for houses, and 99,9% of the time there's no differences between 5 and 99 soft potions. I'd suggest to simply offer a few all encompassing options (like, 99 everything, etc.) and then see where players take them. Others might disagree with me on that.

That being said if you want to split SartingInventory from KillShops i don't see any problems for pushing it to beta.

PR #482, opened 2021-02-05
Deep Dungeon

pinkpuff:
Deep Dungeon turns Coneria Castle into a 52-floor procedurally generated dungeon whose treasures and monster encounters both get stronger the deeper you go in. Chaos is at the end of the final floor. There is a Bahamut and a TAIL somewhere, as well as three guaranteed Ribbons.


FF1Blazorizer/Tabs/ExperimentalTab.razor
```diff
@@ -111,3 +111,32 @@
 				[Parameter] public bool noTabLayout { get; set; }
 				[Parameter] public bool IsOpen { get; set; } = true;
 			}
+=======
```
latzio, line 114: Got some conflict markers to fix up.

PR #493, opened 2021-02-08
Separate Gold Scaling

mhn65536:
The new flag allows a separate scaling for all Gold Rewards(treasure and enemy drops).

If enabled it rolls a global factor between the selected min and max. This factor is applied to all gold rewards. Then it applies +-10% random value to each of them.

EXP/Gold Boost is ignored. Progressive Scaling will still be applied.


wildham0:
The double randomization seems to be at odds with how all the other double sliders work: the double slider usually means that the user get to set the range of randomization and with the log scale it should do so within expectations; here all the values will cluster on a point in the range, which can be particularly bad if it does so at one extreme or the other.

I'd suggest either switching to a simple slider if you want to keep a small variation over a fixed value (then the user can expect where the price will cluster) or keep the double slider and simply apply the log scale.

mhn65536:
With the new information, this will be bugged and needs rework.

PR #494, opened 2021-02-08
Add ToFR Exit

highqualityshaun:
Adds a Warp tile to ToFR 1F so you can leave the dungeon without EXIT or WARP
Works with Shorten ToFR


latzio (approved):

FF1Lib/Maps.cs
```diff
@@ -645,6 +646,17 @@ public void EnableLefeinShops(List<Map> maps)
 			lefein[0x05, 0x15] = 0x1A; // Clinic Sign
 			lefein[0x06, 0x15] = 0x65; // Crescent Lake Clinic
 		}
+
+		public void EnableToFRExit(List<Map> maps, bool shortToFR)
+		{
+			// add warp portal to ToFR 1st floor
+			maps[(byte)MapId.TempleOfFiendsRevisited1F][17, 20] = (byte)Tile.PortalWarp;
+			if (shortToFR)
```
latzio, line 654: Actually this probably belongs in the short tofr section, not here.

latzio: Can you add it to the landingArea in ShardHunt.cs if the exit tofr flag is set?

latzio (changes requested):
The short tofr check should be in the ShardHunt.cs file where it edits that map.

latzio (approved):

PR #501, opened 2021-02-13
Added flag to give Melmond a clinic

pinkpuff:
It puts (repairs?) a clinic in Melmond. It uses the Crescent Lake clinic tile.


Steven-Biro:
this will need to make sure the taverns work in melmond, if enabled with this flag. I dont believe it will be much work, it only needs to check the flag and include that clinic if it is on

see:
https://github.com/FiendsOfTheElements/FF1Randomizer/blob/dac8c6b290353651338140e6baf12036b6078af5/FF1Lib/Hacks.cs#L295

PR #505, opened 2021-02-14
Puts a clinic in Melmond

pinkpuff:
Puts a copy of Crescent Lake's clinic in Melmond (I deleted the previous PR by accident ^_^;)


Steven-Biro:
I can, this evening, make this work with tavern mode so you dont need to just disable it. After that I will merge it in

Steven-Biro:
ok @pinkpuff, I made a PR here: https://github.com/pinkpuff/FF1Randomizer/pull/2 

once you land that PR, it should auto update here and then this can be landed

Steven-Biro:
@pinkpuff this PR here was updated and is identical to #515 which you opened to replace it.

PRs just say "I want to merge branch X into branch Y". You can update branch X after a PR is opened. In this specific case, you can update this branch as much as you want and the PR will not require anything from you to be automatically updated, so long as you push your changes on the same branch to your remote. Im going to close PR #515 for you and I should be able to reopen this one.

PR #506, opened 2021-02-14
Added flag that removes/replaces Elfland's Silver Sword

pinkpuff:
Removes Elfland's Silver Sword and can optionally replace it with a long sword and/or move it up to Melmond. Disabled when Shops flag is set.


Steven-Biro:
I would suggest a drop down for this instead of a series of check boxes, since its reasonable to enumerate the various situations people would actually want to use.

This is a similar thing to what we did with npc shuffle, initially you could set every single npc individually, but it was swapped to fetch or main npcs because there were just too many options that dont really get used separately.

PR #514, opened 2021-02-20
Melmond Hospitality

pinkpuff:
Adds a clinic to melmond. I think someone fixed it to work with tavern mode?


pinkpuff:
Wrong branch, use #515

PR #518, opened 2021-02-21
Preliminary formatting changes

Steven-Biro:
This formats all the code in the repo and (eventually) will check formatting and checking that it can be built in circleci before allowing stuff to be merged in


FF1Lib/Dialogs.cs
```diff
@@ -516,6 +521,360 @@ public string FormattedItemName(Item item, bool specialItem = true)
 					case Item.Canoe:
 						formatted = FF1Text.BytesToText(Get(0x2B5D0 + 36, 8)).TrimEnd(' ');
 						break;
+					case Item.None:
+						break;
+					case Item.Lute:
```
wildham0, line 526: That's a bit silly, is there a way to have it not do that with switch, or is the best practice here is to use if/else when we need to switch only a few case?

Steven-Biro: This was because it requires all cases to be covered in a switch. I've manually added exceptions to the rule in places we do this, but I think if/else might be better when we only cover a few cases. 

Though, we can decide as a group if this rule is something we want or not

FF1Lib/ItemLists.cs
```diff
@@ -160,59 +160,219 @@ public static class ItemLists
...
-			if (flags.IncentivizeOxyale ?? false) incentivePool.Add(Item.Oxyale);
-			if (flags.IncentivizeCanal ?? false) incentivePool.Add(Item.Canal);
-			if (flags.IncentivizeHerb ?? false) incentivePool.Add(Item.Herb);
-			if (flags.IncentivizeChime ?? false) incentivePool.Add(Item.Chime);
-			if (flags.IncentivizeXcalber ?? false) incentivePool.Add(Item.Xcalber);
-
-			return incentivePool.Concat(ItemLists.AllQuestItems).Distinct().ToList();
+			List<Item> incentivePool = new();
+
+			if (flags.IncentivizeMasamune ?? false)
```
wildham0, line 165: this might be the only thing that's really bothering me, as this change lowers readability imo, but this might be also a sign that these flags should be handled differently

wildham0:
this looks good to me
one question: what's the criteria for keeping lists' items on a single line or putting them on each of their line? going over everything it seems a bit random what list get expanded that way or not

mikelattanziobluescape:
> this looks good to me
> one question: what's the criteria for keeping lists' items on a single line or putting them on each of their line? going over everything it seems a bit random what list get expanded that way or not

If it's anything like prettier it switches to multiline when it doesn't fit within the line length you specify.

FF1Lib/ExtensiveHints/HintSources/FloorHintsHintSource.cs
```diff
@@ -1,108 +1,108 @@
...
-
-		private List<GeneratedHint> GenerateFloorHints()
-		{
-			List<GeneratedHint> hints = new List<GeneratedHint>();
-
-			HashSet<Item> WowItems = new HashSet<Item>(ItemLists.UberTier.Concat(ItemLists.LegendaryWeaponTier).Concat(ItemLists.LegendaryArmorTier).Concat(ItemLists.AllQuestItems).Distinct());
-			HashSet<Item> MehItems = new HashSet<Item>(ItemLists.RareWeaponTier.Concat(ItemLists.RareArmorTier).Distinct());
-			HashSet<Item> SparkleItems = new HashSet<Item>(ItemLists.AllGoldTreasure);
-
-			var locations = rom.generatedPlacement.Where(p => p is TreasureChest).GroupBy(p => p.MapLocation);
```
mhn65536, line 35: Why is var not allowed here?

FF1Lib/Fun.cs
```diff
@@ -271,20 +271,49 @@ public void ShuffleMusic(MusicShuffle mode, MT19337 rng)
 					//Overwrite beginning of crystal theme with a song that initializes properly but plays no notes
 					Put(0x340C0, Blob.FromHex("FDF805E0D8C7D0C480"));
 
-					List<int> AllSongs = new List<int>
+					List<int> AllSongs = new()
 					{
-						0x7C649, 0x7C6F9, 0x7C75A, 0x7C75B, 0x7C62D, 0x7C75D,
```
mhn65536, line 276: In my opinion, inflating that list is also not ideal.

FF1Lib/Fun.cs
```diff
@@ -415,9 +444,9 @@ public void EnableModernBattlefield()
 			// Update Several 16 Byte LUTS related to cursors
 			for (int i = 0; i < 8; ++i)
 			{
-				Data[0x31F85 + i * 2] = (byte)(xCoord - 0x10);            // X Coord of Character targeting cursor
-				Data[0x31F86 + i * 2] = (byte)(yCoord + 4 + ((i % 4) * 0x1C));  // Y Coord of Character targeting cursor
-				Data[0x31F76 + i * 2] = (byte)(0xA7 + (Math.Min(i, 4) % 4 * 0x10)); // Y Coord of Command Menu cursor (last four are identical)
+				Data[0x31F85 + (i * 2)] = (byte)(xCoord - 0x10);            // X Coord of Character targeting cursor
```
mhn65536, line 447: we don't have to do this, do we?

FF1Lib/ItemLists.cs
mhn65536: I would have hoped that it would at least make one-liners to make the whole thing a little more compact.

FF1Lib/LegendaryShops.cs
```diff
@@ -10,22 +10,19 @@ namespace FF1Lib
...
-
-		MapTileSets MapTileSets;
-		ShopData ShopData;
-		TileSet[] TileSets = new TileSet[8];
-
-		Dictionary<byte, List<byte>> UnusedTilesbyTileSet;
-		Dictionary<string, MagicSpell> Spells;
-
-		List<SpellInfo> SpellInfos;
+		private readonly MT19337 rng;
```
mhn65536, line 13: if you kill my whitespace, im leaving...

Just joking, but in my opinion this goes too far.

FF1Lib/LegendaryShops.cs
```diff
@@ -91,7 +91,10 @@ public void PlaceShops()
 
 			ShopData.StoreData();
 
-			for (int i = 0; i < 8; i++) TileSets[i].StoreData();
+			for (int i = 0; i < 8; i++)
```
mhn65536, line 94: I'd strongly prefer this to be a one-liner. In my opinion the extra lines can make already long code even longer and less readable.

I know a lot of this is personal preference, but at one point you have to place some trust into a programmer to code it in an understandable and readable way.

FF1Lib/LegendaryShops.cs
```diff
@@ -200,11 +218,30 @@ private int[] AllocateSlots()
 
 			int[] result = new int[] { 0, 0, 0, 0, 0 };
 
-			if (flags.LegendaryBlackShop ?? false) result[2] = Allocate(ref slots, 5);//Black
-			if (flags.LegendaryArmorShop ?? false) result[1] = Allocate(ref slots, 5);//Armor
-			if (flags.LegendaryWhiteShop ?? false) result[3] = Allocate(ref slots, 4);//White
-			if (flags.LegendaryWeaponShop ?? false) result[0] = Allocate(ref slots, 3);//Weapon
-			if (flags.LegendaryItemShop ?? false) result[4] = Allocate(ref slots, 3);//Item
+			if (flags.LegendaryBlackShop ?? false)
```
mhn65536, line 221: Thats 5 lines inflated to 24.

mhn65536:
I'd prefer it not do anything the VisualStudio "Format Document" command doesn't. 

In the end, you want me to fix my own code as long as i'm around. If you change my code to the point, where i don't recognize it anymore, it'll severly hamper me.

Besides all theese formatting changes don't make the problems of the code go away. Spaghetti code will still be spaghetti code. And shitty code will at most become pretty shitty code.

But all this does, is ensure that neither you nor i know, what my code does. So as long as i'm around, I'd like to keep my code intact.

FF1Lib/ExtensiveHints/HintSources/FloorHintsHintSource.cs
Steven-Biro: because the current rules chooise IDE0008 over IDE0007

https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/style-rules/ide0007-ide0008

Personally, I do prefer explicit types, but this is all up for discussion

FF1Lib/Fun.cs
Steven-Biro: I agree, this is likely a rule we need to tweak, or we need to add a few exceptions

FF1Lib/Fun.cs
Steven-Biro: this does not change the functionality, it is the result of choosing IDE0048 over IDE0047
https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/style-rules/ide0047-ide0048

basically, the formatter will automatically add parentheses anywhere that might cause confusion, so people later on dont make a mistake when reading the code.

FF1Lib/ItemLists.cs
Steven-Biro: I agree that this makes it less readable, though in most cases cramming as much into one line as possible isnt great imo, so as a general rule I think this is good, but this function likely needs exceptions.

FF1Lib/LegendaryShops.cs
Steven-Biro: The rules are up for tweaking, I'll try to find the rule that enforces no blank lines here and we can discuss it as a group.

FF1Lib/LegendaryShops.cs
Steven-Biro: yeah, this is due the the default rules and is up for discussion, in general I do think oneliners can be needlessly confusing when being more verbose only takes up more line -- its not like we need to ration lines. though I do agree that in specific cases oneliners can be better, as showcased here.

Steven-Biro:
> 
> 
> I'd prefer it not do anything the VisualStudio "Format Document" command doesn't.
> 

This is the result of running `dotnet format` which is put out by the dotnet team, so while its not baked into the editor, its nearly just as standard. I'll look into making "Format Document" in Visual Studio run the dotnet formatter, or see if there is a separate command in Visual Studio that will do it for us.


> In the end, you want me to fix my own code as long as i'm around. If you change my code to the point, where i don't recognize it anymore, it'll severly hamper me.
> 
> Besides all theese formatting changes don't make the problems of the code go away. Spaghetti code will still be spaghetti code. And shitty code will at most become pretty shitty code.
> 
> But all this does, is ensure that neither you nor i know, what my code does. So as long as i'm around, I'd like to keep my code intact.

This should be a one time migration, afterwards all the errors will show up in your editor and it will suggest how to fix it as well as block merging stuff that doesnt meet the style rules we end up agreeing on. Therefore, for everything after this is merged, it will be the exact code as you had it before that code is merged in.

Steven-Biro:
> 
> 
> > this looks good to me
> > one question: what's the criteria for keeping lists' items on a single line or putting them on each of their line? going over everything it seems a bit random what list get expanded that way or not
> 
> If it's anything like prettier it switches to multiline when it doesn't fit within the line length you specify. If we don't explicitly set one somehow it's probably like 120 chars.
> 
> In my experience, forced style is worth the price of nobody being 100% happy with what it looks like, as long as IDE integrations can be tweaked to everybody's liking. I've been a format-on-save dude for a few years now, and just not even thinking about pressing enter or spacing things properly while coding, knowing that just saving once makes everything 'perfect' is a cool thing to get used to.

Unfortunately it appears the dotnet-format does not have prettier-style line breaking,  and what we are seeing is the rule to preserve oneline vs when things are on multi-line, it just breaks it all into its own line. There is a [github issue](https://github.com/dotnet/format/issues/246) opened in mid 2019 which has the most ❤️  reactions and  👍  reactions out of all current issues, so there is demand and we might get that update in the future.

On the topic of line breaking as it currently stands though, we might need to disable the current rules as it does not enforce line length but rather forces you, for each collection, to have everything on one line or everything in mutliple lines, not ideal.

FF1Lib/ExtensiveHints/HintSources/FloorHintsHintSource.cs
wildham0: I like var (I mean, it's less work), but since we're manipulating a lot of lists and arrays explicit types are way clearer at first glance and should make it easier to newcomers to figure out what the randomizer is doing

wildham0:
> On the topic of line breaking as it currently stands though, we might need to disable the current rules as it does not enforce line length but rather forces you, for each collection, to have everything on one line or everything in mutliple lines, not ideal.

There just seems to be a few lists that aren't on only one line but get to keep their formatting, like in dialog.cs
`				(new List<MapLocation> { MapLocation.ConeriaCastleRoom1, MapLocation.ConeriaCastleRoom2, MapLocation.MirageTower1,
					MapLocation.SeaShrine1, MapLocation.SkyPalace1, MapLocation.TempleOfFiends1Room1, MapLocation.TempleOfFiends1Room2,
					MapLocation.TempleOfFiends1Room3, MapLocation.TempleOfFiends1Room4, MapLocation.CastleOrdeals1, MapLocation.TempleOfFiends2,
					MapLocation.EarthCave1, MapLocation.GurguVolcano1, MapLocation.IceCave1, MapLocation.MarshCave1 },`

looking at it, maybe it's not applying the rule to lists inside lists.

FF1Lib/ExtensiveHints/HintSources/FloorHintsHintSource.cs
Steven-Biro: one thing that is nice, is you can program it with var, then run the formatter and it will try to replace it with the correct type (it only failed a couple times for me and that was when I had a config issue resulting in the formatter not understanding what a Blob is)

Steven-Biro:
> 
> 
> > On the topic of line breaking as it currently stands though, we might need to disable the current rules as it does not enforce line length but rather forces you, for each collection, to have everything on one line or everything in mutliple lines, not ideal.
> 
> There just seems to be a few lists that aren't on only one line but get to keep their formatting, like in dialog.cs
> ` (new List<MapLocation> { MapLocation.ConeriaCastleRoom1, MapLocation.ConeriaCastleRoom2, MapLocation.MirageTower1, MapLocation.SeaShrine1, MapLocation.SkyPalace1, MapLocation.TempleOfFiends1Room1, MapLocation.TempleOfFiends1Room2, MapLocation.TempleOfFiends1Room3, MapLocation.TempleOfFiends1Room4, MapLocation.CastleOrdeals1, MapLocation.TempleOfFiends2, MapLocation.EarthCave1, MapLocation.GurguVolcano1, MapLocation.IceCave1, MapLocation.MarshCave1 },`
> 
> looking at it, maybe it's not applying the rule to lists inside lists.

that is also entirely possible, Im open to suggestions on how you think we should handle this, try to follow [one the the rules]( https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/style-rules/formatting-rules#wrap-options) or disable them entirely and looking into a third party line length checker, or disable them and let PR reviewers catch that stuff.

FF1Lib/Fun.cs
mhn65536: but that operator precedence you usually learn around the age of 10

FF1Lib/LegendaryShops.cs
mhn65536: thanks

FF1Lib/ExtensiveHints/HintSources/FloorHintsHintSource.cs
mhn65536: 2 years ago i preferred explicit types as well. But when using linq, things can get out of hands very easily like:

`foreach (IEnumerable<IGrouping<KeyValuePair<string, Tuple<string, string>>> item in list)`

and become really readable.

mhn65536:
I'm still not entirely sure, what this will accomplish. I can see a lot of really badly formatted code in the project yes. But formatting it "nicely" will not make the code that much more readable and maintainable.

I mean, just look at what it did with enemizer. That's now a whooping 4800 lines instead of already unbearable 3750. Can anyone seriously argument, that he now understands what this code does any more than without theese changes? Besides beeing spaghetti code, the current formatting isn't actually bad. It's sure different, but consistent, clear and it conveys the individial constructs. It's just a really long spaghettified code piece that'll take a long time to get into, no matter what formatting.

Or take OverWorldMap. The changes it does, are superficial. A few extra lines here and there. But the algorithm of F/E Shuffle still have the same complexity. 

Also the Formatter doesn't change bad variable and function names. It doesn't change that half our project is a partial class. 

I will find theese formatting restrictions mostly annoying, especially if they prevent merges and commits. But they will neither make the code more readable nor more maintainable. I'm 100% sure of that.

FF1Lib/ItemLists.cs
Entroper: Make a list that maps flags to items and iterate over the list with a simple function.

FF1Lib/ExtensiveHints/HintSources/FloorHintsHintSource.cs
Entroper: Hmm.  I like `var` for pretty much everything except primitive types like `int` and `ushort`.  Usually when you're doing a complicated LINQ statement, you don't really care what the exact type is because you're just going to iterate over it in a sec.  You can always hover to see the type anyway if you need to.

FF1Lib/Fun.cs
Entroper: Is it possible to disable the style checker for one line with a comment?

FF1Lib/LegendaryShops.cs
Entroper: Oh, yeah.  Single blank lines should always be acceptable.  I'm okay with the linter removing multiple blank lines or blank lines before close braces.

Steven-Biro:
This is temporarily on hold as I work on the FFRBot rewrite, but seeing as how most of the team wants standardized styling of some sort, I'll update the rules and redo this in a bit.

It will have scripts that can auto-format for those not using VS (or prefer to use a script) and will introduce a circleci check to ensure that the formatting is ok on future PRs 

This will be a good first step in moving towards a slightly more organized development process including tests for adding to the changelog, adding tooltips, and (maybe, if this has enough traction) eventually running tests on FF1Lib itself

PR #519, opened 2021-02-23
Lifereboot/20210222 titan snack

highqualityshaun:
random Titan snack (fun % flag)

re-used NPChint check logic from multi-intrumental Sara: moved that logic to private function avoid duplicate code

tested Titan snack with Shuffle Astos in cases where Astos replaces Titan; tested with Hint givers in villages,dungeons;


FF1Lib/FF1Rom.cs
```diff
@@ -907,6 +907,8 @@ public void Randomize(Blob seed, Flags flags, Preferences preferences)
 
 			rng = new MT19337(funRngSeed);
 
+			TitanSnack(preferences.TitanSnack, npcdata, rng);
```
wildham0, line 910: we reset rng between each Fun% flag so that people racing the same seed get the same fun% results even if they don't have the same preferences, another rng = new MT19337(funRngSeed); should be inserted between TitanSnack and HurrayDwarf

highqualityshaun: ty. done

PR #530, opened 2021-03-07
StartingLevels

mhn65536:
Allows you to specify the starting levels of your characters


FF1Lib/asm/1B.asm
```diff
@@ -190,7 +190,86 @@ LvlUp_CheckIfMax:
...
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+;;
+;;  AwardStartingEXP  [$8496]
+;;
+;;
+;; Simulates 16 BattleRewards with fixed exp
+;; Heals all characters afterwards
+;; Really bad hack with the lute
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+    LDA item_lute ;AD2160 basically i set bit 7 of item_lute here, because i din't want to use up a byte of sram and i'm pretty sure it's never set again ever.
```
wildham0, line 204: may I suggest using a gameflag here instead of the lute? there's a lot of bits available, and while yes you shouldn't technically get more than one lute so bit 7 should never be set, it feels like this a future trap in the making

I think ideally it should bea separate awardxp routine and then call LvlUp_NoDisplay that exists already, we would sidestep the one needed byte, but it would be a bit more complex than directly using LvlUp_AwardAndUpdateExp, so I see why you did it that way

mhn65536: We can use a gameflag no problem. I just thought it way more likely that a random gameflag be used, than bit 7 of a keyitem. Besides we weren't sure, the existing gameflag code can handle more bits.

I was thinking of compatibility. Reusing the same code inherently takes care of all working bugfix flags, just the same way as a battle lvlup would.

wildham0: I remember when we talked about this, it doesn't seem the game does anything weird with the gameflags (ie assuming the unused bits are zero) so I think its safe to use

mhn65536: any preference for what gameflag bit i should use?

wildham0: bit 7 sounds good for a mechanics gameflag

mhn65536: done

wildham0 (approved):

PR #532, opened 2021-03-08
Adds a dock to Bahamut's island so that airship is not required for class change.

tetron:


tetron:
Just needs a screenshot

tetron:
Added a screenshot for the tooltip, let me know if there are any review comments.

latzio (approved):

PR #542, opened 2021-03-14
Equip menu info

tetron:
ROM patch that extends the "shopinfo" feature so that you can also press select on the weapon and armor equip menus to display the gear stats when in "equip" mode.


tetron:
@wildham0 @latzio I think this is sorted out now, it shouldn't conflict with the section of code that @mhn65536 claimed on discord.

wildham0 (approved):

PR #543, opened 2021-03-14
Lefein dock & Gaia mountain pass

tetron:
Adds map edits for Lefein river dock (so that Lefein can be accessed by ship+canoe) and Gaia mountain pass (so that Gaia can be accessed from the Lefein region).

When the Gaia pass is enabled, it makes Gaia conditionally ship-accessible (depending on the Lefein river flag) and creates a new Lefein walkable region so they can be connected via an entrance shuffle Titan's cave wormhole.


FF1Lib/OverworldMap.cs
```diff
@@ -97,6 +98,18 @@ public OverworldMap(FF1Rom rom, IMapEditFlags flags, Dictionary<Palette, Blob> p
...
+			    MapEditsToApply.Add(LefeinRiverDock);
+			    mapLocationRequirements[MapLocation.Lefein].Add(MapChange.Ship | MapChange.Canal | MapChange.Canoe);
+		        }
+			if ((bool)flags.MapGaiaMountainPass) {
+			    MapEditsToApply.Add(GaiaMountainPass);
+			    if ((bool)flags.MapLefeinRiver) {
+				// If Lefein river dock is on, then Gaia also becomes ship-accessible
+			        mapLocationRequirements[MapLocation.Gaia].Add(MapChange.Ship | MapChange.Canal | MapChange.Canoe);
+			    }
+			    _walkableNodes[WalkableRegion.LefeinRegion] = new List<OverworldTeleportIndex>{OverworldTeleportIndex.Gaia, OverworldTeleportIndex.Lefein };
```
latzio, line 111: Excellent.

latzio (approved):

PR #544, opened 2021-03-14
Add BugfixRender3DigitStats flag.

tetron:
This tweaks the rendering of character stats so that character
stats (STR, AGL, etc) are rendered correctly for values over 99.
Previously they would render as A0-A9, B0-B9, etc.


FF1Lib/asm/0E_9030_FixPrintCharStat.asm
```diff
@@ -0,0 +1,33 @@
+char_index      = $67
+tmp             = $10 ; 16 bytes
+unsram          = $6000  ; $400 bytes
+ch_stats        = unsram + $0100  ; MUST be on page bound.  Each character allowed $40 bytes, so use 00,40,80,C0 to index ch_stats
+PrintNumber_2Digit = $8E66
+PrintNumber_3Digit = $8E70
+
+.ORG $8E04
+
+;;; we don't want to change jump targets but we need to
```
Steven-Biro, line 10: is there that many places this is called? it would probably be best to change what calls this to call the new routine

Steven-Biro (approved):
going to approve it so it can get some testing in beta for a while

We should come back and make this always on eventually

tetron:
I ended up revising this, it turns out there was an unused branch that could be combined to get the space for the extra couple of instructions I needed, so this new patch stays in the PrintCharStat routine and doesn't have to jump anywhere else @Steven-Biro

PR #592, opened 2021-04-21
ExpChests

mhn65536:


mhn65536:
converts a configurable random amount of gold chests into exp chests. Exp reward is rolled with the usual log scaling in a configurable range.

PR #615, opened 2021-05-16
ChestInfo

mhn65536:


wildham0 (commented):
I think this should be its own flag; ShopInfo in shops and the equip menu is pretty out of the way so it's easy to tune out for veterans, which isn't the case for the chest textboxes; or maybe add a preference to hide it?

mhn65536:
done

wildham0 (approved):

PR #617, opened 2021-05-18
If floater is removed, don't include Cardia entrances in E/F shuffle.

tetron:
Shard and loose incentive item placement now filters on accessibility before placement, so if a flag makes a chest completely inaccessible, it won't result in a soft lock.

### DIFF
--- a/FF1Lib/Interfaces.cs
+++ b/FF1Lib/Interfaces.cs
@@ -149,7 +149,7 @@
 		bool? DeepTownsPossible { get; }
 		bool? AllowDeepTowns { get; }
 		bool? AllowUnsafeStartArea { get; }
-		bool? NoFloater { get; }
+		bool? IsFloaterRemoved { get; }
 		bool? MapBahamutCardiaDock { get; }
 	}
 	public interface IVictoryConditionFlags
--- a/FF1Lib/Interfaces.cs
+++ b/FF1Lib/Interfaces.cs
@@ -133,7 +133,7 @@
 		int PriceScaleFactorHigh { get; }
 		bool? ExcludeGoldFromScaling { get; }
 		bool CheapVendorItem { get; }
-		bool EnableExtConsumables { get; }		
+		bool EnableExtConsumables { get; }
 	}
 	public interface IFloorShuffleFlags
 	{
@@ -149,6 +149,8 @@
 		bool? DeepTownsPossible { get; }
 		bool? AllowDeepTowns { get; }
 		bool? AllowUnsafeStartArea { get; }
+		bool? NoFloater { get; }
+		bool? MapBahamutCardiaDock { get; }
 	}
 	public interface IVictoryConditionFlags
 	{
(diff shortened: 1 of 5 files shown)


FF1Lib/OverworldMap.cs
```diff
@@ -367,6 +367,18 @@ public void ShuffleEntrancesAndFloors(MT19337 rng, IFloorShuffleFlags flags)
 					keepers.Add(OverworldTeleportIndex.Cardia5);
 				}
 
+				if ((bool)flags.NoFloater)
```
mhn65536, line 370: If NoFloater is on without BahamutDock and Dragonshoard, there's still going to be orphaned chests. Could be a problem for shard hunt, or could lock players out of high level gear.
Perhaps it'd be better to enforce Dragonshoard and BahamutDock on floeaterless.

mhn65536 (changes requested):

FF1Lib/OverworldMap.cs
tetron: I went with the more general solution, it now checks accessibility when placing loose items and shards, so if the Cardia islands are completely inaccessible, it won't place any incentivized items or shards there.

FF1Lib/Interfaces.cs
```diff
@@ -149,6 +149,8 @@ public interface IFloorShuffleFlags
 		bool? DeepTownsPossible { get; }
 		bool? AllowDeepTowns { get; }
 		bool? AllowUnsafeStartArea { get; }
+		bool? NoFloater { get; }
```
wildham0, line 152: I'd suggest using the internal flag IsFloaterRemoved instead of NoFloater; while it doesn't change anything right now, it might simplify things down the road if other flags start removing the floater

FF1Lib/OverworldMap.cs
wildham0, line 370: idem

FF1Lib/Interfaces.cs
tetron: That doesn't work.  `IsFloaterRemoved` is a field on `IVictoryConditionFlags` which is used by item placement, but the code that checks `NoFloater` is `ShuffleEntrancesAndFloors` which uses `IFloorShuffleFlags`, not `IVictoryConditionFlags`.

tetron: Unless you're saying I should just literally rename the field to ` IsFloaterRemoved` but that doesn't work either because it just passes in the `Flags` object which implements `IFloorShuffleFlags` so the field name has to match.

wildham0: oh, I just meant to replace `NoFloater` with `IsFloaterRemoved` in `IFloorShuffleFlags`, then you get access to it in `ShuffleEntranceAndFloors`, the compiler will only get confused if several interfaces are merged in a single one, but then you can precise the one you're using by putting he target interface in () before like:
`(IFloorShuffleFlags)flags.IsFloaterRemoved`

tetron: Oh, I see now.  I missed that.  Will fix.

wildham0 (approved):

PR #620, opened 2021-05-19
exclusive legendary shops

mhn65536:


mhn65536:
removes the inventory from the legendary shops from chests and normal shops.

wildham0 (approved):

PR #622, opened 2021-05-21
Tetron/nofloater entrance shuffle

tetron:
With NoFloater enabled, pin Titan's west to its vanilla location so you are guaranteed to be able to access the Sarda region via the tunnel.

Location of Titan's east, and what is placed at the Sarda entrance, are still random.


mhn65536 (approved):
Looks workable...

PR #625, opened 2021-05-25
Fix CheckCanTake, MIAB ChestInfo

mhn65536:


FF1Lib/asm/11_8200_TalkRoutines.asm
```diff
@@ -789,7 +771,7 @@ lut_TargetJump_Lo:
  .BYTE <DialogueBox-1, <EnterBattle_R, <LoadBattleCHRPal_R, <ReenterStandardMap-1
 
 
-  .ORG $B100
+  .ORG $B120
```
wildham0, line 774: This is updated to $B120, but the actual TalkInBattle code is still put at $B100 and is still referred to as 2000B1 in the talkroutine, is this an error or is it supposed to be moved for other things?

mhn65536: I changed the .org, to see in the listing if everything fits. So that I don't have to move everything around again.

wildham0 (approved):

PR #627, opened 2021-05-31
Loose Moose

mhn65536:


tetron:
Let me see if I understand the behavior:

* The RewardSourcePicker makes a weighted choice of where to put an item (the standard behavior is uniform)
* The "weights" dictionary persists between Picks and that is on purpose
* non-chests (is this just NPCs?) have higher weights in general -- this seems like a new loose item behavior?  Should it also get a flag?
* Forward placement means that every time an item is placed, all the weights on existing locations are reduced.  This means when a placement opens up new locations, those new locations will have higher weights than the previously available locations
** Because they get reduced by 50% on every iteration it seems like those numbers could get very small?  50 iterations and you run out of precision in a double precision number.  Is that intended?
* Spread placement takes the location that was picked and figures out the dungeon it is located in (does this work for NPCs?) and gets the list of chests and reduces the weights to 25%.  This means when a location is picked, dungeons that don't have an item placed yet have higher weight, and a dungeon that has 2 items placed is much less likely to get a 3rd, etc.

FF1Blazorizer/Tabs/ExperimentalTab.razor
```diff
@@ -10,6 +10,8 @@
 			<h4>Maps</h4>
 			<CheckBox UpdateToolTip="@UpdateToolTipID" Id="sanityCheckerV2" @bind-Value="Flags.SanityCheckerV2">Use SanityChecker V2</CheckBox>
 			<EnumDropDown UpdateToolTip="@UpdateToolTipID" Id="owMapExchange" TItem="OwMapExchanges" IsEnabled="Flags.SanityCheckerV2" @bind-Value="Flags.OwMapExchange">Overworld Map:</EnumDropDown>
+			<CheckBox UpdateToolTip="@UpdateToolTipID" Indent Id="fowrwardLoose" @bind-Value="Flags.LooseItemsForwardPlacement">Forward Loose Placement</CheckBox>
```
tetron, line 13: typo "fowrwardLoose"

FF1Lib/ItemPlacement.cs
```diff
@@ -513,6 +513,8 @@ protected override ItemPlacementResult DoSanePlacement(MT19337 rng, ItemPlacemen
 			{
 				placementFailed = false;
 
+				var balancedPicker = new RewardSourcePicker(0.5, 8.5, _checker);
```
tetron, line 516: Should there be another flag to control whether NPCs have boosted weight vs chests?  and where did the weight of 8.5 come from?

tetron (changes requested):

FF1Lib/RewardSourcePicker.cs
```diff
@@ -0,0 +1,104 @@
...
+			//If Forward placement is active, reduce the weight of all reward sources that were elligible for an item this step.
+			//Normally chest like ToF or Matoya have a chance to roll a loose item at every step, because they are accessible from the start.
+			//TFC is accessible way later so is used in way less rolls.
+			//By reducing the weight on chests, that already had a chance, it's more likely to select a chest or npc from a newly opened up area(hence forward placement).
+			//The forward placement is deactivated for incentive items, so it doesn't skew the placement.
+			sum = 0.0;
+			foreach (var s in sources)
+			{
+				var v = weights[s.Address];
+				if (forward) weights[s.Address] = v * reduction;
```
tetron, line 74: Since you're multiplying by 50% repeatedly you should set a numeric floor on weights, e.g. Math.Min(v * reduction, .0000000001)

Same for spread

FF1Lib/RewardSourcePicker.cs
```diff
@@ -0,0 +1,104 @@
...
+			//The spread placement remains active for incentive items. There is only one incentive location per dungeon and access requirement.
+			//This has the effect of reducing the chance of a loose item in incentivized dungeons.
+			//There is a small chance of two incentive locations in one dungeon with F/E. But the effect will not be noticable by any unaware observer.
+			if (spread)
+			{
+				foreach (var s in checker.GetNearRewardSources(sources, result))
+				{
+					if (weights.TryGetValue(s.Address, out var v))
+					{
+						weights[s.Address] = v * reduction * reduction;
```
tetron, line 96: Math.Max(v * v * reduction, .0000000001)

FF1Blazorizer/Tabs/ExperimentalTab.razor
mhn65536: pk

FF1Lib/ItemPlacement.cs
mhn65536: i have integrated it into the spreadPlacement Flag.

FF1Lib/RewardSourcePicker.cs
mhn65536: The limit is kind of useless. When a chest falls below 1 in 4 billions, it may as well drop out completely.
All chests dropping to low values is well covered by double bevavior.
Should any numerical issues really arise, there is a safety net that doesn't let if fail.

FF1Lib/RewardSourcePicker.cs
mhn65536: The limit is kind of useless. When a chest falls below 1 in 4 billions, it may as well drop out completely.
All chests dropping to low values is well covered by double bevavior.
Should any numerical issues really arise, there is a safety net that doesn't let if fail.

FF1Lib/ItemPlacement.cs
mhn65536: That number is a "tuned" parameter. I divided the number of chests by the number of npcs. Took half of that and looked through some spoiler logs to see if it was too high or too low.
I looked through some spoiler logs again and lowered it a little.

tetron: ok, I suggest taking this reply you just wrote here and putting it in a comment in the code.

FF1Lib/RewardSourcePicker.cs
tetron: ok

FF1Lib/ItemPlacement.cs
mhn65536: ok

tetron:
I still think weighting NPCs should be its own flag separate from spread placement.  I might want to play "oops all loose" with the possibility of multiple items in the same location but still have NPCs/fetch quests less likely to be complete duds.

FF1Blazorizer/wwwroot/tooltips/tooltips.json
```diff
@@ -25,6 +25,24 @@
...
+		"Id": "forwardLoose",
+		"title": "Forward Loose Placement",
+		"screenshot": null,
+		"description": "Tries to place loose items into newly opened up areas. It creates key item chains where one key item leads to the next, drives items out into the last corners of the world and can create some really nice fetch quest chains."
+	},
+	{
+		"Id": "looseNpcBalance",
+		"title": "Loose Npc Balance",
+		"screenshot": null,
+		"description": "Makes it more likely loose items land on Npcs."
```
tetron, line 44: NPC is an acronym (Non-Player Character) so it should be capitalized

FF1Blazorizer/wwwroot/tooltips/tooltips.json
```diff
@@ -25,6 +25,24 @@
...
+	},
+	{
+		"Id": "forwardLoose",
+		"title": "Forward Loose Placement",
+		"screenshot": null,
+		"description": "Tries to place loose items into newly opened up areas. It creates key item chains where one key item leads to the next, drives items out into the last corners of the world and can create some really nice fetch quest chains."
+	},
+	{
+		"Id": "looseNpcBalance",
+		"title": "Loose Npc Balance",
```
tetron, line 42: Capitalize NPC

mhn65536: ok

tetron (approved):

PR #642, opened 2021-06-15
Save preferences and ROM data at a single domain.

tetron:
Workaround for the issue that every instanced site has its own separate local storage, and thus its own separate preferences.

Uses an iframe to load a stub page at a single domain which can then save and load key-value data.  The domain where preferences get stored is determined by where the iframe is loaded from.

done: preferences and the ROM data

todo: convert presets to this scheme.

### DIFF
--- a/FF1Blazorizer/Tabs/FunTab.razor
+++ b/FF1Blazorizer/Tabs/FunTab.razor
@@ -36,13 +36,10 @@
 	[Parameter] public Action<string, MouseEventArgs> UpdateToolTipID { get; set; }
 	[Parameter] public Action<string> SetStatusMessage { get; set; }
 
-	protected override void OnInitialized()
+	protected override async void OnInitialized()
 	{
 		Flags.Preferences = new Preferences();
-		JSRuntime.InvokeVoidAsync("getFFRPreferences",
-		                          "preferences",
-					  DotNetObjectReference.Create(this),
-		                          "LoadPreferences");
+		await LoadPreferences();
 	}
 
 	void SavePreferences()
@@ -52,10 +49,9 @@
 		SetStatusMessage.Invoke("Preferences saved.");
 	}
 
-	[JSInvokable]
-	public void LoadPreferences(string prefs) {
-	        if (prefs == null || prefs == "") return;
-
+	async Task LoadPreferences() {
+		string prefs = await JSRuntime.InvokeAsync<string>("getFFRPreferences", "preferences");
+		if (prefs == "" || prefs == null) return;
 		Flags.Preferences = JsonSerializer.Deserialize<FF1Lib.Preferences>(prefs);
 		if (Flags.Preferences.SpriteSheet != null) {
 		  SpriteSheetMessage = "Using custom sprites.";
--- a/FF1Blazorizer/Tabs/FunTab.razor
+++ b/FF1Blazorizer/Tabs/FunTab.razor
@@ -1,5 +1,6 @@
 ﻿@using FF1Lib;
-@using System.Text.Json;
+@using Newtonsoft.Json;
+@using System.Reflection;
 @inject Blazored.LocalStorage.ILocalStorageService LocalStorage
 @inject IJSRuntime JSRuntime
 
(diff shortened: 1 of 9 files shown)


FF1Blazorizer/Tabs/FunTab.razor
```diff
@@ -37,30 +38,40 @@
...
-		      SpriteSheetMessage = "Using custom sprites.";
+	async Task LoadPreferences() {
+		string prefs = await JSRuntime.InvokeAsync<string>("getFFRPreferences", "preferences");
+		if (prefs == "" || prefs == null) return;
+		Flags.Preferences = JsonSerializer.Deserialize<FF1Lib.Preferences>(prefs);
+		if (Flags.Preferences.SpriteSheet != null) {
+		  SpriteSheetMessage = "Using custom sprites.";
 		}
+		StateHasChanged();
+	}
```
Steven-Biro, line 60: Im concerned about how this might work with changing formats for the preferences. This will be more of an issue with presets, but is still something to consider with preferences.

Can you confirm that this works with:
1. blank preferences
2. renamed preferences
Im not sure if it will throw if `prefs` does not match `FF1Lib.Preferences`

FF1Blazorizer/wwwroot/js/site.js
```diff
@@ -1,168 +1,207 @@
...
+    //}
+    var response = JSON.parse(e.data);
+    if (window.FFRPreferencesCallbacks[response.key]) {
+        var resolve = window.FFRPreferencesCallbacks[response.key];
+        resolve(response.data);
+    }
+};
+
+async function getFFRPreferences(keyname) {
+    var iframe = document.getElementsByTagName('iframe')[0];
```
Steven-Biro, line 193: use an id and grab it by that instead, so this doenst break if another iframe is added before this one.

FF1Blazorizer/wwwroot/index.html
```diff
@@ -49,5 +49,10 @@
     </app>
     <script src="_framework/blazor.webassembly.js" autostart="false"></script>
     <script src="js/site.js"></script>
+
+    <!-- This iframe is a hidden stub page that we will interact with to store preferences at
+         a single domain, to avoid having separate preferences/saved ROM/presets on every
+         instanced site version.  Where we load this from determines that domain. -->
+    <iframe src="preferences.html" height="0" width="0"></iframe>
```
Steven-Biro, line 56: wouldnt this be `<site>.finalfantasyrandomizer.com/preferences.html`, where <site> will be different for each instanced site? How does this allow for sharing localStorage? Or am I missing something obvious?

Steven-Biro (approved):
feel free to land once the comments are addressed/resolved

FF1Blazorizer/wwwroot/index.html
tetron: Yea, that's a mistake, it is supposed to be `src="https://beta.finalfantasyrandomizer.com/preferences.html"`

FF1Blazorizer/wwwroot/js/prefframe.js
```diff
@@ -0,0 +1,23 @@
+window.onmessage = function(e) {
+    /*if (e.origin !== "http://example.com") {
+        return;
+    }*/
```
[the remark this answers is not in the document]
tetron: Removed it.

FF1Blazorizer/wwwroot/js/site.js
tetron: Done

FF1Blazorizer/Tabs/FunTab.razor
tetron: @mhn65536 already did a fix to handle incompatible preset flags.
This does something a little simpler, incompatible preferences fields should be ignored by a no-op error handler, and it merges the newly saved presets with the existing presets to maintain forward compatibility if you save preferences on an older site, it won't lose keys from preferences on a newer site.

FF1Blazorizer/wwwroot/index.html
tetron: Fixed.

PR #649, opened 2021-06-16
move new consumables to item/equipment tab

mhn65536:


wildham0 (approved):

wildham0 (changes requested):
Let's put some screenshot in the tooltips, then we can move it out of beta

wildham0 (approved):

PR #650, opened 2021-06-17
Crop Screen

mhn65536:


FF1Lib/FF1Rom.cs
```diff
@@ -1050,6 +1050,8 @@ public void Randomize(Blob seed, Flags flags, Preferences preferences)
 			talkroutines.WriteRoutines(this);
 			talkroutines.UpdateNPCRoutines(this, npcdata);
 
+			if (flags.TournamentSafe || flags.CropScreen) ActivateCropScreen();
```
wildham0, line 1053: Modify tournamentsafe tooltip so it indicates that it's also turned on with it

wildham0 (approved):

PR #651, opened 2021-06-17
Slow Hacks

mhn65536:
Since we already have a player with motion sensitivity, i think this should be added, so they can play along hte same seeds, but slower.


highqualityshaun (commented):
Not sure this is appropriate for Fun%/Preferences, as it will change the timing of actions between ROMs that have it active or inactive. My understanding is the purpose of Fun%/Preferences is things that will not change the game in a way that will impact the timing. Someone racing with Slow Hacks for transitions opted out will take longer to get through animations and, over the course of an entire seed, will add up to minutes?

mhn65536:
Atm, people that can't play with SpeedHacks on, cannot take part in any community seed(like async weekly) and therefore cannot talk about the seed and the experiences with others. Opting out of certain speed hacks puts them at a disadvantage compared to other players, so it isn't unfair, but it allows them to experience the same monsters and routing and take part in the conversation after the "race".

tetron:
I think this is a good argument.  It's like "renounce autosort".

tetron (approved):

PR #653, opened 2021-06-20
Uninstall the service worker from the front page.

tetron:
Hopefully this will ensure the redirect works.


tetron:
I might have sneaked this one in on another related PR :smirk:

PR #660, opened 2021-06-24
The "New version available" button now works with instanced sites.

tetron:


Steven-Biro:
this probably kills the pwa. 
Also, does it affect the caching of the site? previously you could run it offline after going there once (cause of the pwa stuff). I know some people use this or have incredibly bad internet, like minutes to load a new version of the site level bad, and that caching really helped them a lot.

Steven-Biro:
There might need to be some discussion and a clear decision on how the versioned sites should work with the redirecting and all that, so a larger action plan can be worked on at once, and then broken up into individual things people could work on

tetron:
> this probably kills the pwa.
Also, does it affect the caching of the site? previously you could run it offline after going there once (cause of the pwa stuff). I know some people use this or have incredibly bad internet, like minutes to load a new version of the site level bad, and that caching really helped them a lot.

I have put that all back.  Installing the PWA works again.

> There might need to be some discussion and a clear decision on how the versioned sites should work with the redirecting and all that, so a larger action plan can be worked on at once, and then broken up into individual things people could work on

I did a writeup:

https://github.com/FiendsOfTheElements/FF1Randomizer/blob/deee65aadf62879ee9a9c4a221128f5a832464df/Doc/deployment.md

wildham0 (approved):

PR #665, opened 2021-07-02
Added new starting equipment flags

iwritecode791:
Added a flag for a starter pack of equipment: gives 4 random common weapons, 4 random buyable armors (1 fighter, 1 rm, 2 mage).
Added a flag for a random weakness targeting weapon (not xcal).

### DIFF
--- a/FF1Lib/ItemLists.cs
+++ b/FF1Lib/ItemLists.cs
@@ -35,6 +35,13 @@ namespace FF1Lib
 			Item.Ribbon, Item.Ribbon, Item.OpalGauntlets, Item.OpalGauntlets,
 		};
 
+		public static readonly IReadOnlyList<Item> TypeWeaknessWeapon =
+		new List<Item> {
+			Item.FlameSword, Item.IceSword, Item.DragonSword, Item.GiantSword,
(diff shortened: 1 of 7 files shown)


FF1Lib/ItemLists.cs
```diff
@@ -35,6 +35,13 @@ public static class ItemLists
 			Item.Ribbon, Item.Ribbon, Item.OpalGauntlets, Item.OpalGauntlets,
 		};
 
+		public static readonly IReadOnlyList<Item> TypeWeaknessWeapon =
```
wildham0, line 38: For compatibility with weaponizer, you can simply cycle through all weapons and check the bytes for elemental and type weakness (if it's excalibur, it's $FF, if there's none it's $00)

`
var weapondata = new Weapon(weaponid, rom);
if weapon.ElementalWeakness ­­> $00 & < $FF
if weapon.TypeWeakness ­­> $00 & < $FF
  then it's a weapon that does extra damage, add it to the list
`

iwritecode791: Good call wildham, I've added this kind of logic. It will now be useable with weaponizer.

PR #682, opened 2021-07-16
oops no item magic

mhn65536:
Should have done this a long time ago. Those pesky bane swords and thor hammers are a real thorn in my side ;-)

### DIFF
--- a/FF1Blazorizer/wwwroot/tooltips/tooltips.json
+++ b/FF1Blazorizer/wwwroot/tooltips/tooltips.json
@@ -25,6 +25,12 @@
 	},
 
 
+	{
+		"Id": "noItemMagicCheckBox",
+		"title": "No Item Magic",
+		"screenshot": null,
+		"description": "Enabling this, removes the magic from weapons and armor. The names will still suggest that the items have spells and you can still use them in battle, but they will do nothing."
+	},
 
 	{
 		"Id": "funEnemyNamesCheckBox",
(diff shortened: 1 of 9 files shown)


tetron:
I'm fine with the feature but the implementation needs some polish:

* No item magic should disable "all items have magic" and "insert RUSE item" in the user interface
* The flag should be passed to Weaponizer
** When true, skip spell assignment: https://github.com/FiendsOfTheElements/FF1Randomizer/blob/dev/FF1Lib/Weapon.cs#L466
** When true, set spellIndex == 0xFF (this will override Defense/Thor Hammmer)  https://github.com/FiendsOfTheElements/FF1Randomizer/blob/dev/FF1Lib/Weapon.cs#L479

mhn65536:
I've initially given thees things some thought. Even vanilla there are items that suggest that they cast a spell. Like the Bane Sword. Implementing the feature "correctly" means changing all spellcasting item names, including items that people know should cast spell without their name implying it(Thor hammer). Removing the item command from the battle menu, which will be causing grief to some players, as it messes with their muscle memory.
Disabling or changing flags that generate new spellcasting gear(ruse item, all weapons cast spells, weaponizer).

All theese are hefty changes, that require thorough testing. So i thought to make this flag trolly. The items taunt you that they cast spells. You can find the Thor hammer, but it does nothing. Examining an item with the shop info will still state that it doesn't cast a spell.

FF1Blazorizer/wwwroot/tooltips/tooltips.json
```diff
@@ -25,6 +25,12 @@
 	},
 
 
+	{
+		"Id": "noItemMagicCheckBox",
+		"title": "No Item Magic",
+		"screenshot": null,
+		"description": "Enabling this, removes the magic from weapons and armor. The names will still suggest that the items have spells and you can still use them in battle, but they will do nothing."
```
tetron, line 32: Need to fix the tooltip to indicate spellcasting items are renamed.

tetron (approved):

PR #685, opened 2021-07-17
V2 NoOverworld Support

mhn65536:
NoOverworld Support without fake OverworldMap


FF1Lib/Dialogs.cs
```diff
@@ -112,8 +112,12 @@ public enum newTalkRoutines
 			Talk_GiveItemOnItem,
 			Talk_Astos,
 			Talk_kill,
-			Talk_Chaos
+			Talk_Chaos,
+			Talk_Floater,
```
wildham0, line 116: These are dynamically inserted by NoOverworld (see PrepNPCs() ), these id's might not match the real id's of the routine
so you'd have to either add the code in newTalkRoutines too and then modify PrepNPCs to use them or have PrepNPCs return the right ID for them

wildham0 (approved):

PR #712, opened 2021-09-02
Tetron/procgen map2

tetron:


FF1Lib/MapExchange/OwMapExchange.cs
```diff
@@ -9,6 +9,7 @@
 using System.Reflection;
 using System.ComponentModel;
 using RomUtilities;
+using FF1Lib.Procgen;
```
mhn65536, line 12: can't comment below:

foreach (var df in data.DomainFixups) domains.SwapDomains(df.From, df.To);
foreach (var df in data.DomainUpdates) domains.Data[df.To] = originalDomains.Data[df.From];

Please add some null checks here.

mhn65536 (approved):

FF1Lib/MapExchange/OwMapExchange.cs
tetron: Sure thing.  I worked around it by assigning an empty array to DomainFixups but it would certainly be better to not crash there!

PR #721, opened 2021-09-09
Add some new fun% enemy names

tetron:
WzOgre →Grover (from Sesame Street)
WzVAMP → Edward (from Twilight)
R.GOYLE →ARGYLE (sounds similar)
R.Ankylo →FunPolic (from DW1R)
WzMUMMY →MOMMY (someone has to be in charge)
BADMAN →Batman (I always think this when I see them)
EVILMAN →OKAYMAN (name moved over)

Whether to capitalize names or not is confusing

### DIFF
--- a/FF1Lib/FF1Rom.cs
+++ b/FF1Lib/FF1Rom.cs
@@ -937,7 +937,7 @@ namespace FF1Lib
 
 			if (preferences.FunEnemyNames && !flags.EnemizerEnabled)
 			{
-				FunEnemyNames(preferences.TeamSteak);
+			    FunEnemyNames(preferences.TeamSteak, rng);
 			}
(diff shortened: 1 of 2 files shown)


DocRosalia:
Kary -> Kelly (joke from when I accidentally said Kelly instead of Kary)
WizOgre -> DIRGE (wizogres are usually in formations with cerebuses)

tetron:
I don't think WzOgres run with cereberus, it's the blue one that's weaker, hyena maybe?

I like "Grover" for WzOgre because they are both blue.

I should definitely add Kary -> Kelly.

tetron:
I have been schooled

wildham0 (approved):

PR #752, opened 2021-11-21
Rom stamping

SaggingRufus:
Implemented ROM Stamping feature for debugging purposes.

I have reserved 350 bytes starting at 1B A000 to store the following information about the rom
   - Seed (64 bytes)
   - Encoded Flag String (158 bytes)
   - Git SHA used to generate the ROM, full size reserved (40 Bytes)
   - SHA-1 of the input rom (40 Bytes)
   -  Filler (48 Bytes)
The filler is currently just writting 48 spaces as a reserve incase the ROM stamp needs to be bigger.


FF1Lib/RomStampModel.cs
```diff
@@ -0,0 +1,20 @@
...
+using System.Threading.Tasks;
+using Newtonsoft.Json;
+using Newtonsoft.Json.Schema;
+using System.IO;
+
+namespace FF1Lib
+{
+	public class RomStampModel
+	{
+		public String s { get; set; }
```
latzio, line 14: Now THIS is where comments would be useful.

For example, what each of these letters means, and why they're just one letter (to minimize the JSON size in the ROM).

FF1Lib/RomStampModel.cs
```diff
@@ -0,0 +1,20 @@
...
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using Newtonsoft.Json;
+using Newtonsoft.Json.Schema;
+using System.IO;
+
+namespace FF1Lib
+{
+	public class RomStampModel
```
latzio, line 12: I don't love the name RomStampModel, Why not something like BuildInfo or DebugDigest. 

The class isn't really specific to ROMs. Stamp is somewhat illustrative of the purpose. "Model" at the end isn't super helpful either.

FF1Lib/FF1Rom.cs
```diff
@@ -1507,5 +1514,18 @@ public void ShuffleRng(MT19337 rng)
 			Put(BattleRngOffset, battleRng.SelectMany(blob => blob.ToBytes()).ToArray());
 		}
 
+		private void StampRom(String _seed, String _encodedFlagString, String _commitSha, String _inputRomSha){
+			RomStampModel stamp = new RomStampModel();
```
latzio, line 1518: I _think_ you should be able to initialize this inline which will make it a little nicer: https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/how-to-initialize-objects-by-using-an-object-initializer

FF1Lib/FF1Rom.cs
```diff
@@ -1507,5 +1514,18 @@ public void ShuffleRng(MT19337 rng)
...
+			stamp.s = _seed;
+			stamp.f = _encodedFlagString;
+			stamp.g = _commitSha;
+			stamp.r = _inputRomSha;
+			System.Text.Json.JsonSerializerOptions options = new System.Text.Json.JsonSerializerOptions();
```
latzio, line 1523: Check the default values on these options. I bet you can omit this entirely.

FF1Lib/FF1Rom.cs
```diff
@@ -1183,7 +1187,10 @@ public void Randomize(Blob seed, Flags flags, Preferences preferences)
 
 			if (flags.TournamentSafe || preferences.CropScreen) ActivateCropScreen();
 
-			WriteSeedAndFlags(seed.ToHex(), Flags.EncodeFlagsText(flags));
+			String seedHex = seed.ToHex();
+			String encodedFlagText = Flags.EncodeFlagsText(flags);
+			WriteSeedAndFlags(seedHex, encodedFlagText);
+			StampRom(seedHex, encodedFlagText, FFRVersion.Sha,inputRomHash);
```
latzio, line 1193: Do you need these parameters?

If you do this inside WriteSeedAndFlagsAndDigest I think FFRVersion.Sha and inputRomHash are already in scope. Don't need anything new here at all.

FF1Lib/FF1Rom.cs
```diff
@@ -114,7 +115,10 @@ public void Randomize(Blob seed, Flags flags, Preferences preferences)
 				rng = new MT19337(BitConverter.ToUInt32(hash, 0));
 			}
 			if (flags.TournamentSafe) AssureSafe();
-
+			// Create a SHA-1 String of the original input rom (used in Rom Stamping process later)
+			using (SHA1 hasher = SHA1.Create()){
```
latzio, line 119: Can you not do this inline? Also the comment isn't really necessary since it's very clear what is being done here.

`inputRomHash = SHA1.Create().ComputeHash(Data.ToBytes())

FF1Lib/RomStampModel.cs
latzio: I think my favorite after mulling it about would be `RandomizeDigest` since it's essentially a digest of the result of the call to Randomize.

Then you could define it inline right above the Randomize method if you like.

It's fine in its own file too, but surely you can chuck all those ugly using statements!

latzio (changes requested):
See comments!

FF1Lib/FF1Rom.cs
SaggingRufus: Im not sure I agree with this one, the WriteSeedsAndFlags does a lot of processing, I think adding this here might obscure what that method is already doing

FF1Lib/FF1Rom.cs
```diff
@@ -102,7 +103,23 @@ public static async Task<FF1Rom> CreateAsync(Stream readStream)
 
 			return rom;
 		}
+		public class RandomizerDigest
+		{
+			//This JSON gets embeded into the rom, so the property values have been
+			//shortened to save space
+			public String s { get; set; } //Randomizer Seed
```
mhn65536, line 110: If you look at this link: https://www.newtonsoft.com/json/help/html/jsonpropertyname.htm
There is a simple way to make the PropertyName in teh code and the Name in the json differ.

I'm not pushing for a change here. With the comments it's totally fine by me. But I just wanted to let you know in case you need it sometime.

SaggingRufus: This is more elegant, I like it!

SaggingRufus:
I got distracted and made a mistake with this merge, will fix and comment. 
- [x] git blame my kids

FF1Lib/FF1Rom.cs
SaggingRufus: ok so while I like this, in the middle of trying to make this change visual studio crashed and now I can build the application anymore cause of this
image

I reverted the property changes because I simply cannoy test them at this time

SaggingRufus:
ok, this can be reviewed now XD turns out my Visual Studio is now broken, so I am ok with this for now lol

FF1Lib/FF1Rom.cs
```diff
@@ -1183,7 +1203,13 @@ public void Randomize(Blob seed, Flags flags, Preferences preferences)
 
 			if (flags.TournamentSafe || preferences.CropScreen) ActivateCropScreen();
 
-			WriteSeedAndFlags(seed.ToHex(), Flags.EncodeFlagsText(flags));
+			String seedHex = seed.ToHex();
+			String encodedFlagText = Flags.EncodeFlagsText(flags);
+			WriteSeedAndFlags(seedHex, encodedFlagText);
+
+			Blob romStampPayload = Encoding.ASCII.GetBytes(System.Text.Json.JsonSerializer.Serialize(new RandomizerDigest(seedHex, encodedFlagText, FFRVersion.Sha, inputRomHash)));
```
latzio, line 1210: Did you see my earlier comment about a null terminator? I think you achieve this, at least, by doing `PutInBank(0x1B, 0xA000 + romStampPayload.size(), Blob.FromHex("0x00"));` or something like that. Then you can scan for the null byte which will be a lot better than scanning for a "}"

latzio: Also something like jsonAscii or jsonBlob is just a better name for this in my opinion, but that's _very_ nitpicky and it's no big deal.

FF1R/Program.cs
```diff
@@ -6,10 +6,11 @@
 	using FFR.Common;
 
 	[Command(Name = "ff1r", Description = "Final Fantasy Randomizer"),
-	Subcommand("presets", typeof(Commands.Presets)),
+	 Subcommand("presets", typeof(Commands.Presets)),
 	 Subcommand("generate", typeof(Commands.Generate)),
 	 Subcommand("procgen", typeof(Commands.Procgen)),
-	 Subcommand("render", typeof(Commands.MapRender))]
+	 Subcommand("render", typeof(Commands.MapRender)),
+	 Subcommand("getdigest", typeof(Commands.GetDigest))]
```
latzio, line 13: Again, nitpicky, but I like the sound of just "digest" more than "getdigest"

FF1Lib/FF1Rom.cs
```diff
@@ -102,7 +103,23 @@ public static async Task<FF1Rom> CreateAsync(Stream readStream)
...
+		public class RandomizerDigest
+		{
+			//This JSON gets embeded into the rom, so the property values have been
+			//shortened to save space
+			public String s { get; set; } //Randomizer Seed
+			public String f { get; set; } // Encoded flag string
+			public String g { get; set; } // Git SHA used to generate this seed
+			public String r { get; set; } // SHA-1 of the input rom
 
+			public RandomizerDigest(String _seed, String _flags, String _gitSha, String _romSha)
```
latzio, line 115: This constructor isn't necessary AND since all the types are identical it can be error prone. If somebody gets the order wrong it blows up! I highly recommend reading this article on parameterless constructors. https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/how-to-initialize-objects-by-using-an-object-initializer

latzio (changes requested):
See a few last comments.

FF1Lib/FF1Rom.cs
SaggingRufus: done and done

FF1R/Program.cs
SaggingRufus: done!

FF1Lib/FF1Rom.cs
SaggingRufus: I had no idea this existed! it makes sense.
Done

SaggingRufus:
here we go again!

FF1Lib/FF1Rom.cs
SaggingRufus: got it to work!

FF1Lib/FF1Rom.cs
```diff
@@ -102,7 +106,19 @@ public static async Task<FF1Rom> CreateAsync(Stream readStream)
 
 			return rom;
 		}
-
+		public class RandomizerDigest
+		{
+			//This JSON gets embeded into the rom, so the property values have been
```
latzio, line 111: Please a space between `//` and `This` like you do elsewhere. Also you misspelled embedded, but I do appreciate the comment.

FF1Lib/FF1Rom.cs
```diff
@@ -1183,7 +1202,20 @@ public void Randomize(Blob seed, Flags flags, Preferences preferences)
...
+
+			Blob digestJson = Encoding.ASCII.GetBytes(JsonConvert.SerializeObject(new RandomizerDigest
+			{
+				seed = seedHex, 
+				encodedFlagString = encodedFlagText, 
+				commitSha = FFRVersion.Sha, 
+				inputRomSha1 = inputRomHash
+			}));
+			PutInBank(0x1B, 0xA000, digestJson);
+			//It's a little wont, but I couldn't get the blob size. So I made it a string and got the length of that
```
latzio, line 1217: https://docs.microsoft.com/en-us/dotnet/api/microsoft.azure.storage.blob.blobproperties.length?view=azure-dotnet-legacy

latzio: Also isn't the hex a hex string? Wouldn't that be nearly twice as long?

latzio: Pardon me it's not the .NET blob, it's the ROMUtilities one that Entroper provided.
https://github.com/Entroper/RomUtilities/blob/master/RomUtilities/Blob.cs

It still has the property!

SaggingRufus: When I tried this before I was getting compilers, but that was also around the same time that autocomplete wasn't working. So may have just been that the environment wasn't quite set up properly I'll try that again tonight.

PR #763, opened 2021-12-09
Tournament Safe + Resources Pack change the seed

wildham0:


tetron:
I see it changes the seed, it should also change the displayed hash?

wildham0:
oh yeah, good catch
should be fixed now

tetron:
Your could modify flagstext the way I did for maps instead of adding new parameter to WriteSeedAndFlags

Or make the map a parameter add well, just seems like there shouldn't be two ways to do it.

tetron (approved):

PR #791, opened 2022-01-27
Gamescreen Tracking

wildham0:


jtoyoda:
Oh man this would be super helpful to fix the autotracker sometimes getting bad data

PR #810, opened 2022-02-19
Bahamut is scaled like a boss.

tetron:
Swole bahamut gets a nastier slate of spells.

### DIFF
--- a/FF1Lib/Enemies.cs
+++ b/FF1Lib/Enemies.cs
@@ -52,6 +52,7 @@ namespace FF1Lib
 			public const int Phantom = 51;
 			public const int Mancat = 55;
 			public const int Vampire = 60;
+		        public const int Ankylo = 78;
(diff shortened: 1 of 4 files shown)


tetron:
> I haven't tested it, but it looks clean (and trust you when you say you tested it)
> 
> I'd prefer normal Fight Bahamut doesn't grant such generous exp, but it wouldn't be the end of the world if left as is.
> 
> For buffed Bahamut, I was not envisioning every skill would be a nasty one, just most of them. Depending on the script assigned this could be a nonissue.
> 
> Change 8000 base exp for normal quality bahamut to 3000? (between Kary1 and Kraken1?)

I dropped normal Bahamut XP down to 3000.

I diluted buff Bahamut's skills a little bit by adding adding CREMATE (lower damage AoE), SNORTING and STINGER to the pool.

highqualityshaun (approved):

PR #817, opened 2022-03-04
rdungeon

mhn65536:


highqualityshaun (commented):
<TreatWarningsAsErrors>false</TreatWarningsAsErrors> in FF1Lib.csproj

not sure if intentional or a temp change you forgot to change back. seems wrong though

PR #822, opened 2022-03-14
Clinics heal all statuses

pinkpuff:


wildham0:
how does this interact with tavern mode? is it compatible?

pinkpuff:
> how does this interact with tavern mode? is it compatible?

I don't know off hand, and I don't know how to check that. It modifies the usual clinic routine to clear all statuses and only set HP to 1 if the character was dead. It also changes the part that checks who to list in the clinic menu by checking if they have ANY statuses rather than if they specifically have the "dead" status.

wildham0:
PubReplaceClinic() in hacks.cs, it seems to be called after so either it doesn't change that part of the code and it's fine or it will overwrite and your flag won't work, you'll have to add some alternate code in pubreplaceclinic

mhn65536 (approved):

tetron:
Should we merge this?  It's too bad if it is incompatible with tavern recruitment but that's not a good enough reason to keep it stuck in pull request limbo forever.  Is there anything else we're waiting on?

PR #823, opened 2022-03-15
Clinics don't cost money

pinkpuff:


FF1Lib/FF1Rom.cs
```diff
@@ -973,6 +973,18 @@ public void Randomize(Blob seed, Flags flags, Preferences preferences)
 				ItemsText[(int)Item.Ribbon] = ItemsText[(int)Item.Ribbon].Remove(7);
 			    }
 
+			if (flags.FreeClinic)
+			{
+				var pointers = Get(ShopPointerOffset, ShopPointerCount * ShopPointerSize).ToUShorts();
```
mhn65536, line 978: Please pass your flag into the ScalePrices method at line 1038. Then you can change the price in Scale.cs line 148 and following.

Also does this set the price of inns to 0 as well? If it does, is that intended?

PR #879, opened 2022-06-27
Added Random Trap Tile Option + Cleanup

iwritecode791:
Added an option to choose from all 256 monster formations (minus bosses) rather than just the shuffled pool or just the b-side pool.

Made all the different tile options an enum dropdown rather than a series of incompatible checkboxes.


iwritecode791:
Got a few more ideas... will pr again when complete.

PR #887, opened 2022-07-07
Fix swapped Early/Late Lockpicking mistake, Tooltip Stuff, Tweak QoL flag ordering

DarkTwinge:


highqualityshaun:
please remove the rip ankylo change to the Fight Bahamut tooltip. while true, it's part of the community knowledge where if someone asks they can find this out, leaving it out is preferred in my opinion (less is more). but mostly, you're changing a tooltip that is already being changed in another PR (fight bahamut minions) and it'll create a conflict

DarkTwinge:
> please remove the rip ankylo change to the Fight Bahamut tooltip. while true, it's part of the community knowledge where if someone asks they can find this out, leaving it out is preferred in my opinion (less is more). but mostly, you're changing a tooltip that is already being changed in another PR (fight bahamut minions) and it'll create a conflict

I definitely like detailed/accurate info where possible, but this one is very minor and I'm not bothered by its removal - went ahead and pushed a revert.

PR #891, opened 2022-07-11
.NET 6

Entroper:
Just updates everything to .NET 6. No AOT compilation for Blazor yet.

I did demo a few things like global usings, file-scoped namespaces, and top-level Program.cs, but I didn't apply that everywhere.


Entroper:
Some CSS is messed up, likely because the <app> tag is gone (it's now just a <div id="app">).  Don't merge until we fix.

PR #917, opened 2022-08-26
Split Encounter Sliders

DarkTwinge:
Ocean Encounters is now its own Slider
Sliders are now based directly on the actual encounter rate they correlate to rather than a multiplier


FF1Lib/Scale.cs
```diff
@@ -440,30 +440,15 @@ public void SetProgressiveScaleMode(Flags flags)
 			PutInBank(0x1B, 0x806D, Blob.FromHex("20CBCFEAEAEAEAEA"));
 		}
 
-		private void ScaleEncounterRate(double overworldMultiplier, double dungeonMultiplier)
+		
+		private void ScaleEncounterRate(double LandEncounterRate, double SeaEncounterRate, double DungeonEncounterRate)
 		{
-			if (overworldMultiplier == 0)
-			{
-				PutInBank(0x1F, 0xC50E, Blob.FromHex("EAEA"));
```
tetron, line 447: You want to keep the code patches -- I believe just setting a value of 0 means you get 1/256 encounters.

PR #928, opened 2022-09-04
Revert Alt Fiends graphics (temp)

wildham0:
Revert alt fiends graphics until imagesharp's compatibility issues are fixed

### DIFF
--- a/FF1Lib/Enemies.cs
+++ b/FF1Lib/Enemies.cs
@@ -1261,6 +1261,13 @@ namespace FF1Lib
 
 			var assembly = System.Reflection.Assembly.GetExecutingAssembly();
 
+
+			// remove when image handling is fixed
+			alternateFiendsList.Shuffle(rng);
+			await Task.Delay(1);
+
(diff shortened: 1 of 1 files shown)


tetron:
I just made PR #933 to revert to .net 5 to work around this issue.

Aside from not wanting to see features removed, I'm not confident that that the ImageSharp issues are limited to alt fiends, it's just as likely that it also crashes on hero sprite import, terrain tile import, etc.  In which case we'd have to disable all of those features.

tetron:
You mentioned that maybe it isn't even an ImageSharp issue at all, it's an async function issue?

wildham0:
Seems so to me, but i don't know enough about async programming to be sure, the crashing happens after the new sprites are written to memory, so looks like one of the thread doesn't return correctly that the task is done (or is trying to return into the wrong thread [i just made that up]).

It would also explain why custom sprites import hasn't been affected; while it's also async it doesn't have the same complexity as importfiends.

PR #938, opened 2022-09-13
 Class Builder: Transmooglifier (beta) 

Tempus:
Added a full class randomizer. There's a checkbox in the experimental beta feature tab. I have a ClassesTab that includes the proper exclusions as well, but this one doesn't have it since I didn't want to change non-beta things.

Highly recommend turning off class specific hacks, blursings, and Chaos mode in the Classes tab. Also not compatible with custom sprite replacements (well, it works it'll just be non-sensical).

Features: Adds procedurally generated heuristically sound classes with custom graphics to the game. Press select to view details.

Updated the pull request to include the custom icon hacks. This allows anyone to create and use custom font tiles, there are about 0x40 left.


FF1Lib/Classes.cs
```diff
@@ -340,39 +340,177 @@ public void SetMpGainOnMaxGain(Flags flags, FF1Rom rom)
...
+			if (!(bool)flags.RandomizeClass && !(bool)flags.Transmooglifier && !(bool)flags.RandomizeClassChaos)
 				return;
-			}
 
-			// Strings to build info screen in game
+			RandomizeClassHacks(flags, rom);
+
+			if ((bool)flags.Transmooglifier)
+				Transmooglify(flags, rng, rom);
+			else if ((bool)flags.RandomizeClass)
```
mhn65536, line 350: RandomizeClassChaos can only be activated if RandomizeClass is. As such you'll have to swap the branches here. Currently Chaos classes doesn't work anymore.

Tempus:
Sorry, in my local I have everything placed and with mutual exclusions in the classes tab, like this: 
image

However I was told to not change tabs and to just put mine in the Beta tab so that's what I did. Can you advise further?

FF1Blazorizer/wwwroot/tooltips/tooltips.json
```diff
@@ -1524,7 +1524,13 @@
 		"Id": "RandomizeClassChaosCheckBox",
 		"title": "Chaos Mode",
 		"screenshot": "RandomizeClassChaosCheckBox.png",
-		"description": "Instead of simply applying \"basic\" blessings and curses to classes, Chaos Mode shuffles all the stats, spellcasting abilities, and equipment permissions between all the classes, making for very different classes from the vanilla game. Only for experienced players that are comfortable with Blursed Classes in general."
+		"description": "Chaos Mode shuffles all the stats, spellcasting abilities, and equipment permissions between all the classes, making for classes with very different abilities from the vanilla game. Only for experienced players."
```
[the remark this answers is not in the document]
Tempus: No, I found the original tooltip to be a little bit editorial in nature, and also didn't fit as well when there was a third class randomization option, so I trimmed it down a bit for clarity and tone. Not particularly attached to this if you prefer the previous description.

FF1Lib/Classes.cs
Tempus: Sorry, in my local I have everything placed and with mutual exclusions in the classes tab, like this:
image



However I was told to not change tabs and to just put mine in the Beta tab so that's what I did. Can you advise further?

FF1Lib/Classes.cs
```diff
@@ -62,10 +62,10 @@ public enum MpGainOnMaxGain
 
 	public class GameClasses
 	{
-		private List<ClassData> _classes;
-		private GearPermissions _weaponPermissions;
-		private GearPermissions _armorPermissions;
-		private SpellPermissions _spellPermissions;
+		public List<ClassData> _classes;
```
wildham0, line 65: don't make these public, they're only there for the internals of gameclasses and they can't be trusted if they're exposed, you should use ArmorPermissions, WeaponPermissions, SpellPermissions and ClassData directly

FF1Lib/Classes.cs
mhn65536: just fix the if statement, so that it works with the UI the way it is. somewhat like RandomizeClasses && !ChaosClasses

Tempus: But that's not how the code works? It did not work that way to begin with. Chaos class and Transmooglifier are both mutally exclusive with Blursings, not depending on Blursings being active. The previous UI was simply wrong - if you activated Chaos Mode no blursings would take effect, they were completely different code branches that Chaos Mode hijacked to borrow the dataScreen. I abstracted dataScreen out so I could use it too, and the updated UI correctly reflects their existing relationship in the live build and in the one in the PR.

Tempus: I guess to be clearer:

- If you activated Blursings you got Blursings.
- If you then clicked Chaos Mode, blursings did nothing and Chaos Mode was active
- Likewise, the new feature Transmooglifier also does not activate Blursings

All three are mutally exclusive Class Randomization options, and the new UI reflects this accurately, the previous one from current dev branch was mistaken.

PR #951, opened 2022-09-25
Make the Randomized Shard names a Fun% checkbox

DarkTwinge:


highqualityshaun:
titan snack is not a checkbox, renaming its ID to titanSnackCheckbox seems odd

this "opt in" to fun shards will confuse players and commentators for a while

DarkTwinge:
> titan snack is not a checkbox, renaming its ID to titanSnackCheckbox seems odd

Whoops, yeah was just trying to do some quick cleanup on the IDs while I was there and made it *more* confusing instead of less - fixed.

PR #955, opened 2022-10-15
Tetron/procgen dungeon

tetron:


tetron:
Reported on discord:

> I tried to generate a seed with both generated earth cave and generated waterfall, but it can't satisfy incentive placement doing that. Turning off waterfall and keeping earth cave proc gen worked.

> There did seem to b e a lot of linked or otherwise empty chests

highqualityshaun (approved):

PR #959, opened 2022-11-01
Add Compatibility with non-vanilla Spell Lists

HermitDreams:
### ExtConsumables.cs ###
- Wraps the contents of AddExtConsumables() in "value is not None"
- This avoids needing to search for CURE, PURE, and SOFT in the spell list, even when no consumables were added
- Added searches for those spell types, but they are not used yet
  - **FindWeakCure()** Looks for HP Up, One Ally, and Power less than 33
  - **FindAntidote()** Looks for Remove Ailment with the Poison bit on
  - **FindLotion()** Looks for Remove Ailment with the Stone bit on

### BugFixes.cs ###
- Altered the LOK2, HEL2, and SABR checks to make sure specific vanilla bytes are present before patching them
  - **LOK2** confirms its contents are Evade Up Targeting All Enemies before changing the routine
  - **HEL2** confirms its in-battle graphic is Orange Stars before changing the effect
  - **SABR** confirms it is Effect 0x10, non-elemental, and self-targeting before changing the accuracy

All proposed changes have been tested against both vanilla and altered spell lists and behave appropriately.
The minimalist approach to the conditions is to limit how much of the input file the randomizer should know about.

LOK2 is probably fine, but any concerns about HEL2 and SABR can be addressed.

My reasoning for HEL2 was that my generator patches spells before randomization, and never assigns Color 0x27 to Shape 0xC0, so they would never collide.

SABR might collide with Apply Fire Resist to Self, but the acc change won't affect the spell meaningfully.
It won't collide with Remove Stun from Self because you cannot physically perform that action in battle.
If it happens to roll completely vanilla SABR in that slot (would require I *don't* have level balance on, as SABR hits +37 at Level 7), well, it earned that +10 to Hit.

### DIFF
--- a/FF1Lib/BugFixes.cs
+++ b/FF1Lib/BugFixes.cs
@@ -96,8 +96,8 @@ namespace FF1Lib
 		public void FixSpellBugs()
 		{
 			Put(0x33A4E, Blob.FromHex("F017EA")); // LOCK routine
-			Put(0x3029C, Blob.FromHex("0E")); // LOK2 spell effect
-			Put(0x302F9, Blob.FromHex("18")); // HEL2 effectivity
+			if (Get(0x3029B, 2) == Blob.FromHex("0110")) Put(0x3029C, Blob.FromHex("0E")); // LOK2 spell effect
+			if (Get(0x302FD, 2) == Blob.FromHex("C027")) Put(0x302F9, Blob.FromHex("18")); // HEL2 effectivity
 
 			// TMPR and SABR
 			// Remove jump to PrepareEnemyMagAttack
@@ -111,7 +111,7 @@ namespace FF1Lib
 			// Call new saving code from BtlMag_SavePlayerDefenderStats
 			Put(0x337C5, Blob.FromHex("209FB6EAEAEAEA"));
 			// SABR's hit% bonus
-			Put(0x30390, Blob.FromHex("0A"));
+			if (Get(0x30391, 3) == Blob.FromHex("100004")) Put(0x30390, Blob.FromHex("0A"));
 
 			TameExitAndWarpBoss();
 
--- a/FF1Lib/ExtConsumables.cs
+++ b/FF1Lib/ExtConsumables.cs
@@ -46,47 +46,50 @@ namespace FF1Lib
 
 		public void AddExtConsumables()
 		{
-			ChangeItemJumpTable();
-
-			WriteOutOfBattleRoutines();
-
-			WriteDrawItemBoxEndOfItemIndex();
-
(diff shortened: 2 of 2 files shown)


highqualityshaun:
While I can appreciate a check before applying a put, there's no consequent warning or error that will appear in the UI that says "I tried to do this but then didn't". In other words, if someone has the Spell Fixes option on (very common), there is minor potential for only some of the spell fixes to be applied, in the event any of these checks fail? I'm not sure it's a giant concern, it's just something we don't often bother with (Puts with no checks are rampant in the rando code with a lot of faith that they coexist gracefully)

Additionally, you're adding some methods that won't actually be called by the randomizer, but you're anticipating maybe it will be something useful in the future? In a sentence, do you have plans for these methods or additional features? what's the motivation?

highqualityshaun:
Also, while it is a valid check, having a check on the color of a spell is a little unusual to me. Simply a never been done before (most spell checks for helpers are based on behavior)

HermitDreams:
> While I can appreciate a check before applying a put, there's no consequent warning or error that will appear in the UI that says "I tried to do this but then didn't". In other words, if someone has the Spell Fixes option on (very common), there is minor potential for only some of the spell fixes to be applied, in the event any of these checks fail? I'm not sure it's a giant concern, it's just something we don't often bother with (Puts with no checks are rampant in the rando code with a lot of faith that they coexist gracefully)

Good catch. I should have something that at least writes to console and says whether each rule passed or failed (and based on what criteria). Might be a while before I dig back in and write it up because I haven't been keeping myself, uh, sharp.

> Additionally, you're adding some methods that won't actually be called by the randomizer, but you're anticipating maybe it will be something useful in the future? In a sentence, do you have plans for these methods or additional features? what's the motivation?

This one is in case Cure, Pure, and Soft potions end up needing to be assigned the way ExtConsumables currently does it, but without assuming the spell list is in vanilla order. (Last I tested, the potions were working as intended without running through that portion of ExtConsumables, but I still wrote them up because all the additional potions use the same method)

> Also, while it is a valid check, having a check on the color of a spell is a little unusual to me. Simply a never been done before (most spell checks for helpers are based on behavior)

It was the method that had the least chance of giving a false positive, because there are no other vanilla spells with that combination of colour and shape, and anything my personal generator comes up with won't make that combination, either.

mhn65536:
The code that you put under that if, also does some additional battle loop changes. For example Life in battle requires some of the modifications.
It would be much safer to fix the search instead of disabling a lot of code. Like if you don't have ExtConsumables enabled, yield the original spell idex of the heal potion.

PR #963, opened 2022-11-23
'Power' Item Tweaks & Other Misc.

DarkTwinge:
+ "Titan's Trove" map edit renamed "Move Titan"
+ 'Guaranteed' Items are now named "Power Rod/Gauntlet Effect" to be clearer
+ Power Gaunt now has both Vanilla and None settings; spots in the code which incorrectly thought "None" meant "No effect" have been fixed
+ Power Staff can now also be set to FOG2
+ "Random" for Power Effects now only selects 'None' 25% of the time (instead of 50%)
+ Grayed out several options more appropriately based on which settings are selected
+ Clarified a handful of tooltips and added a new one to explain which enemies actually count as "Bosses"
+ Fixed a minor gray-out mistake if Transmog was set to tri-state
+ Fixed a few typos


highqualityshaun:
I'm not sure about the change from the UI about "Guaranteed Offense Item" / "Guaranteed Defense Item" to "Power Gauntlet Effect" / "Power Staff Effect" -- because after the changes, the items are no longer named PowerS or PowerG, they are appropriately renamed, example RUSE Staff, SABR Gauntlet.

DarkTwinge:
> I'm not sure about the change from the UI about "Guaranteed Offense Item" / "Guaranteed Defense Item" to "Power Gauntlet Effect" / "Power Staff Effect" -- because after the changes, the items are no longer named PowerS or PowerG, they are appropriately renamed, example RUSE Staff, SABR Gauntlet.

These are the changed labels I proposed when I was working with this stuff and wildham liked them. The current name is confusing with the 'guaranteed' nomenclature, and I think referring to the actual item affected is more clear than the generic "defense/offense" phrasing even tho the in-game name changes when non-vanilla. Did you have something else in mind?

highqualityshaun:
I actually never thought about how "Guaranteed" could be confusing. It's not guaranteed to be found. it just guarantees there is an item magic of a certain type. reconsidering this with your comment, you're right. thanks

highqualityshaun:
closed after confirming with DarkTwinge (base is too different, PR from 2 yrs ago no longer relevant)

PR #964, opened 2022-11-30
Added Lockpicking support to Transmooglifier

Tempus:
It changes the class check on lockpicking to match a Transmooglifier genned class. The chosen class will have Lockpicking in the description.

It chooses a classed based on the following:
1. In order, it will prefer the following thematic lockpickers* if available: THIEF, PIRATE, MARAUDER, BERSERKR, FENCER, JUGGLER, SOLDIER, PUGILIST, Bl.BELT 
2. If none are available, it will pick any character with over 35 agility level buffs
3. If none are available, it will pick any character with over 35 strength level buffs
4. If none, it'll pick whoever is in original Thief slot

_* Some pick locks with their fists_


Tempus:
Rebasing this to 4.7.0

PR #973, opened 2023-01-21
Generate Overworld Map Spoiler Image

mhn65536:


FF1Lib/OverworldMap.cs
```diff
@@ -1690,61 +1696,119 @@ public void ShuffleObjectiveNPCs(MT19337 rng)
...
+				tp.TilePropFunc |= TilePropFunc.OWTP_SPEC_CHIME;
+				tileset.TileProperties[i] = tp;
+			}
+
+			tileset.StoreData();
 		}
 
-		tileset.StoreData();
-	    }
+		public void RenderMap()
```
tetron, line 1756: I don't know if you were looking at the code in `FF1Randomizer/FF1R/Commands/Progen.cs:RenderMap()` -- I see here you are not producing full-size maps so it wouldn't be exactly the same.  Would it make sense to render it full size and then shrink it down to 1024x1024 or whatever the target size is?  It looks like you are turning all the 16x16 tiles into 4x4 tiles and then rendering from that.

Another idea would be to deliberately render maps that looks like the in-game map.

mhn65536: Most image operations are stupidly slow in the browser, so i tuned it to a bearable level. Thus I copied your code and modified it to increase the performance in the browser to reasonable levels. Things that can'T be done in the browser:
- encoding the image into anything other than bmp takes 40+ seconds(just the encoding part).
- drawing images in the transform takes 20+ seconds for all tiles..

Don't get me wrong, the way you did it is the preferable and cleaner way, but it unfortunately doesn't work in the browser. Basically I tried around a bit and that was the only thing I could get to work with reasonable performance.

I can increase the size, but the bmp is already 3MB in size, increasing it to 2048x2048 would make it 12MB. It would take roughly 6 instead of 2 seconds to render. The javascript engine takes a couple of seconds extra to download12MB in base64, but the FileSize would still be a bit ridiculous.

PR #980, opened 2023-02-03
Allow Flame and Ice armor to prevent damage from ice and lava tiles

markyys:
When the Flame armor is equipped the equipping character will not take any damage from ice tiles in Ice cave, and when Ice armor is equipped the equipping character will not take any damage from lava tiles in Volcano.

This is only useful with "Lethal Damage Tiles" on, and probably only when the damage per tile has been increased significantly.

I did not put much thought into the name of this setting, the name of the flag describing it, or the toolip for it.

There's probably a smarter way to do some of this. Sorry about the one big commit, but I wasn't really sure how to break it down since it's all so interrelated.

I have tested what I believe is all combinations of the relevant flags, all of which are working as I anticipate:
- 
- `Armor Resists Damage Tiles` without `Lethal Damage Tiles`, no adjustment to tile damage.
- `Armor Resists Damage Tiles` with `Lethal Damage Tiles`, no adjustment to tile damage.
- `Armor Resists Damage Tiles` without `Lethal Damage Tiles`, tile damage increased.
- `Armor Resists Damage Tiles` with `Lethal Damage Tiles`,  tile damage increased.
- `Lethal Damage Tiles`,  without `Armor Resists Damage Tiles`, no adjustment to tile damage.
- `Lethal Damage Tiles`,  without `Armor Resists Damage Tiles`, tile damage increased.


mhn65536:
The asm most likely does what it claims to do. I'd prefery if the additional code would be always written to the rom. That way we only have to deal with 1 variant. It's written in a way that allows disabling the additional flag stuff anyway.
It's incompatible with armorcrafter though. 7 is not going to be a flame armor with armorcrafter. So it should be disabled for that.

It'S a question whether the flag actually makes a gameplay difference. It's nice for fluff. But in a racing situation, you're going to volcano with or without flame armor. Add to that that you can only protect 1 character.
I'd consider adding at least ice shield and flame shield. Or going after the resistances. There's a variable ch_resist in the party memory for that.

FF1Lib/asm/1E_B100_DamageTilesKill.asm
```diff
@@ -35,20 +40,60 @@ Loop:
...
+		CMP #$22
+		BEQ @Volcano
+		CMP #$23
+		BEQ @Volcano
+		CMP #$24
+		BEQ @Volcano
+		JMP @IceCheck
+
+	@Volcano:
+		LDA #$08 ; Ice Armor ID
```
mhn65536, line 62: You could make your check code simpler by looking for an equipped Ice armor #$88 instead of checking all equipped armor pieces to be an ice armor.

mhn65536:
Does the cur_map reset if you go back to the overworld? There's a flag that makes marsh and dessert tiles do damage on the overworld. if cur_map doesn't reset, you wouldn't take damage after leaving ice cave if you have an ice armor.
Cany ou check for that?

markyys:
> The asm most likely does what it claims to do. I'd prefery if the additional code would be always written to the rom. That way we only have to deal with 1 variant. It's written in a way that allows disabling the additional flag stuff anyway. It's incompatible with armorcrafter though. 7 is not going to be a flame armor with armorcrafter. So it should be disabled for that.

I can add a check for the armorcrafter flag.

> It'S a question whether the flag actually makes a gameplay difference. It's nice for fluff. But in a racing situation, you're going to volcano with or without flame armor. Add to that that you can only protect 1 character. I'd consider adding at least ice shield and flame shield. Or going after the resistances. There's a variable ch_resist in the party memory for that.

The original idea was for only the armor, explicitly not protecting you if you had resist from any other source (including the shields). I don't have a horse in this race, I just thought it was a neat idea and wanted to see if I could implement it.

If the considered opinion is that it should look at "fire resist" or "ice/flame armor and shield" I can make that change.

FF1Lib/asm/1E_B100_DamageTilesKill.asm
markyys: Sure, that makes sense.

markyys:
> > Does the cur_map reset if you go back to the overworld? There's a flag that makes marsh and dessert tiles do damage on the overworld. if cur_map doesn't reset, you wouldn't take damage after leaving ice cave if you have an ice armor.

Good catch: it does not clear cur_map when going to the overworld. Any idea how we can distinguish "on overworld/not on overworld"?

markyys:
It looks like elsewhere in the code they are able to rely on
```
LDA mapflags
LSR A
BCC whatever
```
to tell if we're on the overworld. I think that should work in this case.

markyys:
> The asm most likely does what it claims to do. I'd prefery if the additional code would be always written to the rom. That way we only have to deal with 1 variant. It's written in a way that allows disabling the additional flag stuff anyway.

I did it this way because that's what the existing DesertOfDeath code does, so I have to account for it being there or not being there. I agree that it's easier if the code is always in the ROM, and we just flip a flag from the randomizer, but I didn't want to make that sweeping of a change.

mhn65536:
It's your call about Flame Armor, Shield, Resistances. You can always change your mind later. If you want more opinions, just ask on the dev discord.

PR #982, opened 2023-02-05
LockRespondRate: Add a slider which controls the respond rate.

markyys:
Instead we replace the value at the point where it is read. Instead of reading from the respond rate byte we read a hardcoded value which is written by the randomizer.


highqualityshaun:
I liked that the default slider option is 8 which will pair with the default lock respond true. If any new FFR users think the battle text is too fast we can direct them to QoL to let them slow it down.

I liked that even if I did the "fast hard reset with a left/right input" mash to the new game screen, and that screen now showed "Respond Rate 1" it was still the slider setting "in the background" that was being selected.

What I didn't like, and might cause some confusion, is if the user legit picks something other than 1, the new game screen still says Respond Rate 8 (even though a different respond rate exists in the code)

Please modify so that the new game screen indicates a locked respond rate that matches the slider

markyys:
> Please modify so that the new game screen indicates a locked respond rate that matches the slider

Good point. Done and done.

highqualityshaun (approved):

PR #987, opened 2023-02-24
Add class blursing "Friendship Rune"

markyys:
This is inspired by, but not directly modeled on, the Kindness and Friendship runes in Suikoden 2.

Also included are some unrelated fixes which I did at the same time, and arguably should be in a different PR. Let me know if you want me to split those out.

tl;dr: The more NPCs you talk to, the more melee damage you deal.

This is implemented by using a bit in the game flags (bit 3) to record for each NPC whether you have spoken to them or not. I couldn't find a good place to record this in the C# code, so it's not as discoverable as I would like. Suggestions welcome.
 
The player will be able to use a setting at seed creation time to control how large the damage bonus is. We wil take the number of NPCs spoken to and divide it zero times, one time, or two times, so that the damage is equal to all of, half of, or one quarter of the number of NPCs spoken to.
The default is 1:2, which I think is probably the sweet spot. 1:4 feels a bit weak, and 1:1 feels a bit too strong. This bears testing.


highqualityshaun:
minor tweaks like you did for the variable names, file names, use of a variable instead of 0xXX is fine to include with a feature, no need to separate it out.

on the surface this looks ok I'll wait for some other dev feedback though since the ASM is a little cloudy for me. I don't understand why large chunks of previous code were removed, but I'm guessing you found a more efficient way to do the same thing. Still planning to wait

It's also always a little funny to me the way some new features skip the "experimental" tab, but it makes sense for this to live by the class blursings the same way transmoog did. It's fine where it is, I just like it when brand new features can be highlighted in the experimental section as "the latest stuff"

as for some feedback on the blessing itself, i was expecting it to be party-wide for some reason, if that class was in your party. it seems odd to me that this damage blessing would be on a mage class at all. seems like a "womp womp" if it lands in the pool on a mage, no?

markyys:
> on the surface this looks ok I'll wait for some other dev feedback though since the ASM is a little cloudy for me. I don't understand why large chunks of previous code were removed, but I'm guessing you found a more efficient way to do the same thing.

Yeah, that's basically it. Not really important, just something I did while I was in there.
I was initially unsure if I would need the extra room to fit my new code; in the end I don't think I did.

> It's also always a little funny to me the way some new features skip the "experimental" tab,

Honestly I didn't even realize there was an experimental tab. Could add it there instgead if you like.

>  it seems odd to me that this damage blessing would be on a mage class at all. seems like a "womp womp" if it lands in the pool on a mage, no?

Maybe, but if you have the damage at 1:1 you could be getting up to 200 (easily 50) bonus damage on that mage, which would make then pretty viable as melee.

FF1Lib/asm/1B_B000_StartWithRoutines.asm
```diff
@@ -94,8 +96,18 @@ NotANone2:
...
+  RTS
+
+IsEquipped:
+	@Loop:
+		LDA (btl_ob_charstat_ptr), Y
+		BPL @NoEquip
+			RTS
+	@NoEquip:
+		INY
+		CPY #$20
```
wildham0, line 108: might cause false positives with weapons

FF1Lib/asm/1B_B000_StartWithRoutines.asm
```diff
@@ -301,17 +282,39 @@ Hunter:
...
+				LSR
+				STA tmp
+				DEC npc_damage_divider
+				JMP @divide
+		:
+		STX npc_damage_divider ; restore the divider
+		LDY #$09 ; damage
+		LDA (btl_ib_charstat_ptr), Y
+		CLC
+		ADC tmp
```
wildham0, line 313: no overflow check?

markyys: Good point,  I didn't think of that. Would that just be a check of the appropriate flag, then clamp to 255 if it's set?

FF1Lib/asm/1B_B000_StartWithRoutines.asm
markyys: Good point, I can add a better check.

wildham0: well it's more that if you're doing weapons you're also cycling armors, and then because the game use relative ids for weapons/armors, it might trigger on equipped armors if it's the same id as a weapon

markyys:
I came up with a couple alternative versions of IsEquipped, I'm not sure which one is the most clear.

```
IsEquipped:
	ORA #$80
	STA tmp
	@Loop:
		CPY #$20
		BEQ @NoEquip
			LDA (btl_ob_charstat_ptr), Y
			CMP tmp
			BEQ @Done
				INY
				BNE @Loop ; We know this can never fail to branch, and JMP is more bytes than BNE
	@NoEquip:
		LDA #1
	@Done:
		RTS

IsEquipped:
	ORA #$80
	STA tmp
	@Loop:
		CPY #$20
		BEQ @NoEquip
			LDA (btl_ob_charstat_ptr), Y
			INY
			CMP tmp
			BNE @Loop
			RTS
	@NoEquip:
		LDA #1
		RTS
```

PR #995, opened 2023-04-13
Split "Common Items Can Have Powers" into "Spells" and "Powers"

sandman-ahoy:
Split "Common Items Can Have Powers" (which used to control spells and special abilities) into "Spells" and "Powers".


wildham0:
Sorry for the delayed response. Unfortunately the rando changed significantly since this was posted, so this pr will be closed. If you're still interested, you can submit an updated pr and it will receive a fair review.

PR #996, opened 2023-04-13
Block caster items from the shop in Weaponizer

sandman-ahoy:
- prevents magic items from being in shops in weaponizer, unless Allow Caster & Elite Gear is checked


wildham0:
Sorry for the delayed response. Unfortunately the rando changed significantly since this was posted, so this pr will be closed. If you're still interested, you can submit an updated pr and it will receive a fair review.

PR #1012, opened 2023-06-08
Unarmed LUTs for Absorb and Unequip/Reequip

Tempus:
Re-request of https://github.com/FiendsOfTheElements/FF1Randomizer/pull/1007

I have also added some videos of the test case script. The modified code executes only when:
- Leveling up
- Opening the Armor or Weapon Equip screen
- Closing the Armor or Weapon Equip screen

The intended effect is:
- Black Belts gain Absorb == Level when no armor is equipped (Excluding on first launch, this is a vanilla bug)
- Damage == Level * 2 when no weapons are equipped

The abridged script is as follows:
- Load into a game
- Check all status to verify initial status
- Open/Close Weapon and Armor screens.
- Check all status. If BB Absorb == 1 and no other stats change, it is a success.
- Level Up.
- Check all status. If BB Absorb == Lvl and BB Damage == Lvl*2, and no other class equals those stats, it is a success.

- Repeat above, but equip/unequip gear during step 3 to verify it functions with gear

- Repeat above, but use Transmooglifier. Any character using Unarmed icon (including BB) is valid as the BB for this check. They should not be in the original BB character slot (slot 02).

Videos of above testing:
[TestVideos(AVI).zip](https://github.com/FiendsOfTheElements/FF1Randomizer/files/11692535/TestVideos.AVI.zip)


Tempus:
Moved to pull against dev

PR #1066, opened 2024-04-27
Fixed General Fiend error

Oslodo:
Found error with skill/spells settings.


Oslodo:
Also changed one fiend's weakness.

PR #1068, opened 2024-05-05
Class Magic Additions

Oslodo:
- Pink Mage added as an option for a flipping of the Red Mage/Wizard spell slot permissions
- Black Knight added as an option for changing the Knights magic from White to Black and with Ninja permissions
- Black Knight Keep permissions added to keep the Knight's normal white magic permissions but with black magic spell slots
- White Ninja added as an option for changing the Ninja's magic from Black to White with Ninja permissions
- White Ninja Keep permissions added to keep the Ninja's normal black magic permissions but with white magic spell slots
- Knight level 4 spells slot added as an option to give the knight a random level 4 spell slot in white magic or, if the flags for black knight and keep permissions are turned on, will give a random level 4 spell slot in black magic instead.


Oslodo:
Code cleaned up.

PR #1072, opened 2024-05-23
Fixing Knight Level 4 Spell

Oslodo:
- Added Level 4 spell charges for Black Knight and for the Level 4 Random spell.
- Tooltips updated to reflect this as well.


FF1Lib/Classes/ClassesData.cs
```diff
@@ -385,13 +397,12 @@ public void WhiteNinja(Flags flags)
...
 			}
 
-			if((bool)flags.WhiteNinjaKeep)
+			if ((bool)flags.WhiteNinjaKeep)
 			{
 
 				var blackNinja = _spellPermissions[Classes.Ninja].Where(s => ((int)s % 8) >= 4).ToList();
 				var whitetoblackN = blackNinja.Select(s => (s - 4)).ToList();
 				_spellPermissions.ClearPermissions(Classes.Ninja);
-				_spellPermissions.AddPermissionsRange(whitetoblackN.Select(x => (Classes.Ninja, x)).ToList());
```
wildham0, line 394: did you mean to delete this?

FF1Lib/Classes/ClassesData.cs
```diff
@@ -349,27 +349,39 @@ public void BlackKnight(Flags flags)
...
 
 				var whiteslotsK = _spellPermissions[Classes.Knight].Where(s => ((int)s % 8) < 4).ToList();
-			
 				var whitetoblackslotsK = whiteslotsK.Select(s => (s + 4)).ToList();
-
 				var knightslots = whitetoblackslotsK.ToList();
 				_spellPermissions.ClearPermissions(Classes.Knight);
 				_spellPermissions.AddPermissionsRange(knightslots.Select(x => (Classes.Knight, x)).ToList());
-
+				var KnightLvl4Charge = new List<byte> { 0b00000000, 0b00000000, 0b00000000, 0b00000000, 0b00000000, 0b00000000, 0b00000000, 0b00000000, 0b00000000,
```
wildham0, line 356: so the same array is defined 3 times here, it should defined only once outside the function (but inside the classesdata class), something like
`private static List<byte> KnightLvl4Charge = new() { 0b00000000, etc. };`

FF1Lib/Classes/ClassesData.cs
```diff
@@ -349,27 +349,39 @@ public void BlackKnight(Flags flags)
...
+														0b00000000, 0b00000000, 0b00000000, 0b00000000, 0b00001111, 0b00000000, 0b00001111, 0b00000000, 0b00001111,
+														0b00000000, 0b00001111, 0b00000000, 0b00001111, 0b00000000, 0b00001111, 0b00000000, 0b00001111, 0b00000000,
+														0b00001111, 0b00000000, 0b00001111, 0b00000000, 0b00001111, 0b00000000, 0b00001111, 0b00000000, 0b00001111,
+														0b00000000, 0b00001111, 0b00000000, 0b00001111, 0b00000000, 0b00001111, 0b00000000, 0b00001111, 0b00000000,
+														0b00001111, 0b00000000, 0b00001111, 0b00000000 };
+				_classes[(int)Classes.Knight].SpCGrowth = KnightLvl4Charge;
```
wildham0, line 362: not a big deal, but since you do the assignment in both `if` brackets anyway, you could do it only once outside of them

wildham0 (changes requested):

PR #1079, opened 2024-06-24
Titan Food Updated

Oslodo:
Added new Foods to Titan list


highqualityshaun (changes requested):
IS/ARE(relating to plural form), you have "ENCHLDAS IS CHEESY" and I never expected to review something like this and make you change it but I'm gonna

Oslodo:
Ah, the flying copy paste error, will change.

Oslodo (commented):
Changed requested line.

PR #1085, opened 2024-07-18
Remove Boss Scripts

Oslodo:
- Added the option to remove the fiend scripts
- Added tool tip and image for the above code
- Fixed type in Alt Fiend tooltip back end code, non-user impacting.


BrianCumminger (approved):
Looks good

PR #1098, opened 2024-09-21
Make "Starting AoE Caster Item" guarantee that an AoE attack magic item exists (if possible)

markyys:
I think this will work, but it's hard for me to test. The scenario is "we generated a random set of spells for item magic, none of which happened to include a spell in the 'AoE attack' set". This code should, in that case, pick a random AoE attack spell and add it to the list to be applied to items.

It's arguable that if it does so it should also remove a random spell from the list so that the total number of items with magic does not change.

### DIFF
--- a/FF1Lib/ItemMagic.cs
+++ b/FF1Lib/ItemMagic.cs
@@ -82,6 +82,16 @@ namespace FF1Lib
 				Spells = GetAllSpells(rng);
 			}
 
+			if ((bool)flags.StartingEquipmentRandomAoe && flags.ItemMagicMode != ItemMagicMode.None) {
+				// guarantee that at least one item has an AoE attack spell
+				if (!Spells.Any(spell => spell.IsAoEAttack())) {
+					var spellHelper = new SpellHelper(this);
+					var aoe_spells = new List<MagicSpell>(spellHelper.GetAoEAttackSpells().Select(s => s.Info).ToList());
+					aoe_spells.Shuffle(rng);
+					Spells.Add(aoe_spells.First());
+				}
+			}
+
 			foreach (var item in Spells.Zip(magicItems, (s, i) => new { Spell = s, Item = i }))
 			{
 				WriteItemSpellData(item.Spell, item.Item);
--- a/FF1Lib/Magic.cs
+++ b/FF1Lib/Magic.cs
@@ -1,4 +1,4 @@
-﻿using System.ComponentModel;
+using System.ComponentModel;
 using System.Reflection;
 using Newtonsoft.Json;
 using Newtonsoft.Json.Converters;
@@ -9,7 +9,7 @@ namespace FF1Lib
 
 	public enum SpellRoutine : byte
 	{
-	        None = 0,
+		None = 0,
 		Damage = 0x01,
 		DamageUndead = 0x02,
 		Heal = 0x07,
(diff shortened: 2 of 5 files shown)


FF1Lib/ItemMagic.cs
```diff
@@ -82,6 +82,16 @@ public void ShuffleItemMagic(MT19337 rng, Flags flags)
 				Spells = GetAllSpells(rng);
 			}
 
+			if ((bool)flags.StartingEquipmentRandomAoe && flags.ItemMagicMode != ItemMagicMode.None) {
```
wildham0, line 85: so that doesn't seem to do what is intended, the aoe spell is placed at the end of the list so it will never be picked, and because it doesn't discriminate it could introduce back excluded aoe spells (like nuke in balanced)

FF1Lib/Magic.cs
```diff
@@ -1,1242 +1,1258 @@
...
-		{
-			var fields = typeof(SpellSlotStructure).GetFields(BindingFlags.Public | BindingFlags.Static);
-			return fields.Where(f => f.FieldType == typeof(SpellSlotInfo))
-				.Select(f => f.GetValue(null) as SpellSlotInfo)
-				.Where(t => t != null)
-				.ToList();
-		}
-	}
-}
+using System.ComponentModel;
```
wildham0, line 1: seems formatting was accidentally changed for this file, you should revert it and add back your code

wildham0 (commented):
You can test your change by generating several seeds or placing breakpoints in the code you added. You could add a temporary loop to the feature so it keep running and you can make sure the result is always what's intended.

mhn65536:
I'll explain how the code works. If you randomize item magic and for example pick "support" magic as the pool, then there will be 42(including duplicates) Spells in the pool. The pool is shuffled. There are 12 items that want magic. So the first 12 Spells from the shuffled pool will be assigned to the items.
There's now some possibilities what can happen:
- There is an aoe spell within the first 12 slots(we're good)
- There is an aoe spell in the pool, but not in the first 12 slots(we should pick an aoe spell from the pool and randomly insert it into one of the first 12 slots)
- There isn't an aoe spell in the pool(good question what to do, but very unlikely to happen as all pools allow for aoe inflict status(like sleep or charm))

In case there really isn't and aoe spell in the pool, I'd personally fail with a message(to reroll). It can only happen in combination with spellcrafter, probably will never happen and if it does a new seed will fix it.

wildham0:
i'm pretty sure spellcrafter has guarantee aoe elemental spells anyway, so there's no scenario where the spellbook has no aoe spells

mhn65536:
Yes, but the Support Pool does not include InflictDamage spells, only InflictStatus and Buffs. But Spellcrafter is pretty much guaranteed to generate at least one AoE InflictStatus.

markyys:
This has definitely happened in the while, though it probably is quite rare

FF1Lib/ItemMagic.cs
markyys: That's a good point. Is there an existing "exclude spells" utility which could be used to filter the list before picking a new random spell?
And yes, I did think that we may be required to also remove a spell from the pool, good to know.

wildham0: as madmartin said you don't need to create a new list, the generated list has already all the valid spells, you need to find the aoe's in that list and move it up

FF1Lib/Magic.cs
wildham0: if you want to update formatting you should do it in a separate pr, mixing it up makes it difficult to see what changed

markyys: The changes are in separate commits. If you want to see the non-whitespace changes, look at the individual commits.  That's a good idea in any case, to see what changed and why at each step.

PR #1099, opened 2024-10-31
Add Vehicles to Sprite Sheets

brbrofsvl:
This adds support for custom vehicle sprites to player sprite sheets, with backwards compatibility for previous player sprite sheets.

Guidelines:
1) Sprite sheets with dimensions 208x240 will be processed as before.

2) Sprite sheets with dimensions 208x288 are now supported, with vehicle sprites at the bottom.

3) Each vehicle has 6 16x16 sprites; two each for movement up, down, and left (movement right is just mirrored as with mapmans).

4) The order from top to bottom is ship, airship, canoe, the same order as in the ROM.

5) Vehicles must share a single palette, with magenta processed as transparent.

6) This will require later update to the wiki for more documentation, including the following reference png. I've added a couple of others to demo the feature.

Reference png:
sprites_vehicles

The same with a purple palette:
sprites_vehicles_purple

A sheet replacing canoe with a steamboat:
sprites_steamboat


brbrofsvl:
I did not include sync versions of the vehicle import function, since those only seem to be used by the transmooglifier routines to import transmoog sprites.

wildham0:
what happen if the vehicles have different palettes? (i know it's not supported in-game, but how do the importer deal with it?)

brbrofsvl:
Same as if a mapman head (or body) for a given class has different palettes. If I'm understanding the code correctly, it will print a warning, but since the actual encoding for PPU is bitwise, it can process the image with the first four unique colors. See https://github.com/brbrofsvl/FF1Randomizer/blob/fd1e119f392a4460336b7b99790da34238cc0715/FF1Lib/Sprites.cs#L58 

The one difference between how it imports mapmans and vehicles is that it processes mapmans one at a time, but it processes the vehicles all at once. If I can figure out a way to support alternate canoe sprites, it will be an easy fix -- just import canoe stuff first, and then the other vehicles.

PR #1102, opened 2024-11-19
Adds "true prng" encounter table to Experimental

brbrofsvl:
This adds a "true prng" encounter table to the Experimental tab. Overwrites the subroutine that reads from the fixed 256-step encounter table with one that draws from a 16-bit xorshift prng. The rando seeds this prng and chooses one of 4 related prngs, so that it almost never the same with new seeds.

The rng state is stored in sram as soon as party select is confirmed, and persists across hard/soft resets.


brbrofsvl:
Thanks!

PR #1103, opened 2024-11-20
Adds Alt. poison sfx and disable airship sfx

brbrofsvl:
This adds alternate poison status sfx to QOL. Pretty self-explanatory. No .asm needed for disabling airship sfx because it's changing only a few bytes to bypass the sfx.

### DIFF
--- a/FF1Lib/QualityOfLife.cs
+++ b/FF1Lib/QualityOfLife.cs
@@ -1,6 +1,7 @@
 ﻿using RomUtilities;
 using System;
 using System.Collections.Generic;
+using System.ComponentModel;
 using System.IO.Compression;
 using System.Linq;
 using System.Text;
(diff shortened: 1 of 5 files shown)


FF1Lib/QualityOfLife.cs
```diff
@@ -45,6 +77,75 @@ public void DisableDamageTileSFX()
...
+			// of operations, playing a new sound (or none) once every 4 steps instead of once every frame
+			PutInBank(0x1F, 0xC7FB, Blob.FromHex($"A534F00160A000A200BD0161C903D01FA001BD0B61D007BD0A61C9029011BD0A6138E9019D0A61BD0B61E9009D0B618A186940AAD0D398F01FADA0602903D018{SFXCommand}60EAEAEAEAEAEAEAEAEAEAEAEAEA"));
+		}
+
+		public void DisableAirshipSFX()
+		{
+			// this is very simple, so no .asm saved for this routine
+		
+			// NOP out the sfx during the airship transition animations
+			PutInBank(0x1F,0xE215,Blob.FromHex("60EAEAEAEAEAEAEAEAEAEAEAEAEAEAEA"));
```
wildham0, line 144: how does it interact with airboat?

brbrofsvl: If you need confirmation in the code, airboat goes through the vanilla animation and movement routines once you initiate flight/landing, so skipping the sfx in those doesn't alter any other behavior

https://github.com/FiendsOfTheElements/FF1Randomizer/blob/4fb5e0ca29f0f63ed9c22fe4e67c090cf20e33ea/FF1Lib/asm/1B_A000_AirBoatRoutines.asm#L217

PR #1108, opened 2024-12-12
Print class blursings to ROM generation window

brbrofsvl:
Prints the class blursings to the ROM generation window (and preserved in the filetab). User must click to expand, so that by default it doesn't take up too much space. This saves the necessity of opening the ROM in the blursesspoilers page. Since blurse spoilers are allowed for tournaments, it's coded as a Preference so that it doesn't change the rando RNG. I also added a bit of formatting (bold, indents) for readability.

TODO:
Eventually, blursing spoilers should pull class names from the ROM instead of hard-coding them, for compatibility with transmoog. I opted not to do this now.


brbrofsvl:
The formatting is in crude html -- <b></b> and &nbsp; -- I didn't want to change any CSS or make things too unwieldy with inline styles, spans, etc.; I'd be glad to do it a different way if you'd rather.

PR #1115, opened 2025-02-25
Tri state flag output

1nf3rna:
I added a feature to output the collapsed tristates flags to file and well as a button to apply them to the UI.
I was looking for a way to know what the tristate flags become after generation.


1nf3rna:
Updated to move to a checkbox that shows/hides the reveal button

PR #1121, opened 2025-05-10
Thief Hit% buff feature has been expanded... to give flag crafters more …

FranklinFFR:
…to give flag crafters more fine-tuned control. The Red Mage now also has hit% options and thus can be made into a viable melee, or nerfed to have the melee capablilities of the wm and bm.

A point of consideration is that with the ridiculous hit% growth, if there is a plus hit% blessing, then the two might cause overflow, and I don't know how that would work


highqualityshaun (commented):
modifying the .csproj to include a PNG is unnecessary, and not consistent with the other ways we add images to tooltips.
remove the changes to FF1BBlazorizer.csproj (itemgroup), and just get in the habit of adding the .png or .gif to the Blazorizer/wwwroot/images/tooltips directory and naming it appropriately (we tend to match the filename on id, so it should be redMageHitDropDown.png, reference that in the tooltips.json)

the .png itself is an uncommon size. we've standardized the tooltips to be 256x224 pixels (the output of a screenshot from FCEUX), but yours is 421x342

some mistakes: you've created an enum for ThiefHit but not a corresponding enum of RedMageHit. not sure if you forgot to commit everything you changed for this PR or if your branch actually has a missing bind from the frontend to the backend.
in the same theme, you initialize the RedMageHitBuff to the ThiefHit.Vanilla value, and do a logic comparison incorrectly if (flags.RedMageHitBuff == ThiefHit.Vanilla)

regarding cleanup, you've effectively replaced the ThiefHitRate bool value, but you've left that in the code. you need to remove the ThiefHitRate bool since you've added the ThiefHit enum
simiarly, you've renamed the id from the original thiefHitCheckBox to thiefHitDropDown but left the tooltip image as thiefHitCheckbox.gif

finally, some nitpicking. you've ordered the dropdown values as 2, 4, 3, 1, 5 and it makes more sense to order them numerically as 1, 2, 3, 4, 5. by setting the default value to Double it will mimic the current behavior of the randomizer (defaults to doubling the Thief Hit Rate)
more nitpicking, the variable name "HitBalenced" should be "HitBalanced" or you should move to a more consistent naming convention. you alternate between English descriptions "Useless" "Double" "Ridiculous" and mathematical descriptions "+50%", and the enum descriptions and values are not a match. "Useless" versus "HitNothing"
final nitpicking, it's a writer's challenge to make tooltips that are both informative and exercise brevity (say the thing in as few words as possible). instead of saying:

"Changes the hit rate of Red Mages and Wizards from +2 per level to +N per level, making them more viable as melee. (Or, you can make them as useless at melee as the white and black mages)"

try instead to say:

"Changes the hit rate of Red Mages and Wizards from +2 per level to +N per level, making them more or less viable as melee."

for consistency, I notice the thief tooltip still says "Increases" but you have added an option to decrease it too, so make sure to mark that as "Changes" instead of "Increases"

FranklinFFR:
I have polished the code. Thanks for reviewing it. A couple questions:

Do I misunderstand how enums work in C#? Since the Thief Hit enum has the same contents as the Red Mage Hit enum it seems that the only reason to define both would be for ease of readability. 

The other question is where is the outdated thiefHit bool? I can't find it, and I thought I fully removed it already. I suppose it is possible that I accidentally fixed that after pushing, but that seems unlikely.

highqualityshaun:
re: enums, you have them separate, even though their values are identical, because you want to be able to control the thief's critical hit rate and the red mage's critical hit rates independently. if you only had one, when you changed the thief's critical hit rate, you were simultaneously changing the red mage's critical hit rate.

re: the ThiefHit bool you need to remove, it's in FF1Lib/Flags/Flags.cs
public bool ThiefHitRate { get; set; } = false;
make sure any references to it are removed (JSON presets)

I haven't looked closely at how it appears in the UI, but the ordering of the enums are 1 2 4 3 5. This bothers me probably more than it should. Similarly, the cheeky comment should go

PR #1122, opened 2025-05-10
Lefein shortcut

FranklinFFR:
This flag is pretty simple. It adds two bridges to Lefien that make a massive difference to QoL. The hint giver is now worth checking, and the slab turn-in is now painless. Tooltip is included.


highqualityshaun (commented):
as discussed in the developer discord, your PRs are building on each other. use unique branches per feature so that your PRs only have commits relating to the new feature in them. since these are relatively small features, it might be best to save off the relevant code for the feature, revert your changes to the latest dev, branch off of dev for the individual feature, and recommit your code changes and resubmit the PRs. i only say this because the first feature, thief & red mage hit rate adjustments, requires additional work before it can be accepted

PR #1124, opened 2025-05-16
QoL magic shop menu update

1nf3rna:
added a QoL flag to change the magic shops to function like the weapon and armor shops.
Select Spell > Pay > Select Character


FranklinFFR:
Cool Idea! I think I would like using this flag

1nf3rna:
removed the tristate flag stuff, since that's a different pull request

PR #1129, opened 2025-06-28
Fountain funpercent redo

FranklinFFR:
I cleaned up the code so that it is ready to be merged. I have not managed to come up with any additional text, however. 

I like talking to the fountain but I would like examining the fountain more if some randomness is involved.
ps: wash your face


FranklinFFR:
I think I accidentally forgot to make a new branch again :(
I don't know of an easy way to fix it; I think the easiest option is to make any potential changes for both of them in this pull request so they can be merged at the same time.

FranklinFFR:
Realised I forgot to clarify, this merge has the fountain fun% and also the Lefien shortcut.

